Issue #52, opened 2024-01-02
Mark SKIP/SKIP100/SKIP300/ENDSKIP and WTMULT as unsupported keywords


blattms:
In release 2023.10 the simulator will still stop if these keywords are used.

The old manual still states that SKIP/SKIP100/SKIP300/ENDSKIP are ignored which is wrong. For WTMULT is says that it claims that it is supported.

Best resource for checking the behavior of the simulator is probbably [UnsupportedFlowKeywords.cpp (release 2023.10)](https://github.com/OPM/opm-simulators/blob/release/2023.10/opm/simulators/utils/UnsupportedFlowKeywords.cpp), which lists all unsupported keywords, e.g. `{"ENDSKIP", {true, std::nullopt}},` in line 167. The boolean indicates whether the simulator will stop, last parameter is for further information.

blattms:
Closed by #149

PR #148, opened 2024-02-08
Added script to fix ignored keywords

hakonhagland:
See #52 for background.


scripts/src/fodt/fix_ignored.py
```diff
@@ -0,0 +1,166 @@
...
+                if match := re.match(r"""(\s*<text:p\s+.*?>)(.*?)(</text:p>.*$)""", line, re.DOTALL):
+                    start, txt, end = match.groups()
+                    # Remove span tags from txt
+                    txt = re.sub(r"""<text:span .*?>(.*?)</text:span>""", r"\1", txt)
+                    txt = re.sub(r"""<text:span>(.*?)</text:span>""", r"\1", txt)
+                    if "is ignored by OPM Flow and has no effect on the simulation" in txt:
+                        logging.info(f"Found ignored match in {file}:{line_no}.")
+                        txt = self.replace_ignored(txt, critical)
+                        line = start + txt + end
+                        changed = True
```
blattms, line 46: Just one question. There is ugly case with `... simulatio</text:span><text:span text:style-name="T2789">n.</text:span>`. That is probably not covered and needs to be done manually. I guess that is ok, though as we cannot anticipate these kind of formatting things.

blattms (commented):
This looks really cool. Thanks  a lot.

@lisajulia can you have  a look at the code and merge, please?

scripts/src/fodt/fix_ignored.py
hakonhagland: > That is probably not covered and needs to be done manually.

@blattms If you mean the case that there is no space between closing and starting tag: `</text:span><text:span>`, it should also work. I tested it now.

scripts/src/fodt/fix_ignored.py
```diff
@@ -0,0 +1,166 @@
...
+#    "This keyword is not supported by OPM Flow but would change the results "
+#       if supported so the simulation will be stopped."
+#
+#  - Replacement phrase for non-critical keywords:
+#     "This keyword is not supported by OPM Flow but has no effect on the results so will be ignored."
+#
+@click.command()
+@ClickOptions.maindir(required=False)
+def fix_ignored(maindir: str
+) -> None:
```
lisajulia, line 160: Is the line break here intentional?

hakonhagland: No they are not :) Thanks, for the comment, I will fix this. (I have problems pushing to this branch for some reason, so need to fix this problem first)

hakonhagland: > I have problems pushing to this branch for some reason, so need to fix this problem first

I have updated the PR in a new PR, see #149.

hakonhagland:
Replaced by #149.

PR #149, opened 2024-02-08
Fixed ignored keywords

hakonhagland:
See #52 for background. Used the script in #148 to fixup ignored keywords.


blattms:
> @hakonhagland that looks great. I have scanned through and haven't spotted any unhandled cases. Thank you!

Then my concern in #148 is probably unfounded. That's great!

hakonhagland:
Resubmitted as #152.

PR #147, opened 2024-02-08
Renamed `scripts/` to `scripts/python`

hakonhagland:
For background see #146.

PR #151, opened 2024-02-09
Renamed scripts/ to scripts/python

hakonhagland:
Resubmitting #147 since it was closed and cannot  be reopened

Also rebased this PR.


blattms:
As I have learned from github support. Removing forks and PRs when a repo is going public is intended behavior. :see_no_evil: 
Didn't see that coming...

hakonhagland:
> removing forks and PRs when a repo is going public is intended behavior. Didn't see that coming...

No, that was unexpected to me too.. More info here: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/what-happens-to-forks-when-a-repository-is-deleted-or-changes-visibility

PR #153, opened 2024-02-11
Rewrite docker scripts in Python

hakonhagland:
Instead of using a combination of Bash and PowerShell scripts, this PR rewrites the scripts as a Python project. This will help reducing code duplication between Bash and PowerShell scripts, and also allow for easier extension to more than a single container in the future.

Currently, there is only one Docker image (ubuntu2204-lo7.5.9) but we might extend this to include, for example, an image for a development version of LibreOffice with debugging symbols in the future. This rewrite of the scripts in Python should make it easier to add this in the future.


docker/poetry.lock
```diff
@@ -186,4 +225,4 @@ zstd = ["zstandard (>=0.18.0)"]
 [metadata]
 lock-version = "2.0"
 python-versions = "^3.10"
-content-hash = "4a9fb08e31dcd623e22b077f5847a3d6eb09e88843e639470bf4e3d1216d6496"
+content-hash = "497014acdd1583b7cc8bc8732850e02d34f644444841b6253f845f72328b3241"
```
lisajulia, line 228: @hakonhagland: Do we want to track changes like this? (You added the .lock file to git for [this reason](https://github.com/OPM/opm-reference-manual/pull/108#issuecomment-1914701525), right?)

hakonhagland: @lisajulia Yes that was the reason I added the file to git. I think the content-hash is used to detect changes in pyproject.toml. If that file has changed, poetry needs to recalculate dependencies. So I think it should be tracked by git such that poetry can know when to recalculate dependencies.

blattms:
Only concern was resolved. I interpret this as approval.

PR #156, opened 2024-02-16
Preparations for Developer's Review version.

blattms:
Here I update many things in the main document manually (version, revision, revision history, current author, etc.)

This is in preparation for the pdf to send to some developer's for review.

Addinf a new to the table was a bit tricky as cells in the first column span multiple rows. If you don't increase number. libreoffice will crash and show an IO Error


hakonhagland:
Thanks, can you rebase this PR on main, please?

blattms:
Thanks for noticing. I rebased this now.

parts/main.fodt
```diff
@@ -264148,7 +264148,7 @@ Updated with AFR/TSA Rev-D comments and new  keywords.</dc:description><meta:ini
        <text:p text:style-name="Table_20_Contents"><text:s text:c="2"></text:s>Document Name</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table1883.B22" table:number-columns-spanned="4" office:value-type="string">
-       <text:p text:style-name="P10483"><text:file-name text:display="full">davs://blattms@storage.opm-op.com/remote.php/dav/files/blattms/opm-op/Projects/P20-07 Equinor Maintenance/Reference Manual/2023-04 Manual/OPM_Flow_Reference_Manual_2023-04_Rev-0.odt</text:file-name></text:p>
+       <text:p text:style-name="P10483"><text:file-name text:display="full">OPM_Flow_Reference_Manual_2023-0am_Rev-A.fdot</text:file-name></text:p>
```
hakonhagland, line 264151: This is what I see when I open main.fodt:

image

Where do the text `/home/docker-user/parts/main.fodt` come from in the screenshot above?

blattms: This might always be substituted by the current location. I don't know why. maybe a macro does this? We could just delete this row. It has now real value.

blattms: Maybe we should also do something about the email. Matthew might receive a lot spam because of this.

hakonhagland: > Where does the text `/home/docker-user/parts/main.fodt` come from in the screenshot above?

@blattms I think it could be due to the `<text:file-name>` tag? If select the item and then right-click on it, you can choose "Edit Fields" from the popup menu:

image

Which gives this dialog:

image

hakonhagland: > Maybe we should also do something about the email. Matthew might receive a lot spam because of this.

@blattms  I hope not. @gdfldm Let me know if you want to obfuscate the email, or I can merge this now.

gdfldm: @blattms perhaps it would be better to suggest people raise issues?

blattms: I removed the file-name tag. Thanks a lot for pointing that out @hakonhagland 
I now point to github instead of listing an email,

parts/main.fodt
```diff
@@ -264148,7 +264148,7 @@ Updated with AFR/TSA Rev-D comments and new  keywords.</dc:description><meta:ini
        <text:p text:style-name="Table_20_Contents"><text:s text:c="2"></text:s>Document Name</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table1883.B22" table:number-columns-spanned="4" office:value-type="string">
-       <text:p text:style-name="P10483"><text:file-name text:display="full">davs://blattms@storage.opm-op.com/remote.php/dav/files/blattms/opm-op/Projects/P20-07 Equinor Maintenance/Reference Manual/2023-04 Manual/OPM_Flow_Reference_Manual_2023-04_Rev-0.odt</text:file-name></text:p>
+       <text:p text:style-name="P10483">OPM_Flow_Reference_Manual_2023-04_Rev-A.fdot</text:p>
```
hakonhagland, line 264151: @blattms Thanks for the update. Could we add a space in front of "github", see screenshot below. Also, the document name: should it be `main.fodt`? Or `OPM_Flow_Reference_Manual_2024-02_Rev-A.fodt` (change year to 2024 and fdot to fodt). Also, when I click the "github" link nothing happens.
image

blattms: github is fixed.
I think document name "main.fodt" is not very speaking. I kept the old naming convention here OPM_Reference_Manual_<release-version>.fodt. I will correct the extensio then to have OPM_Reference_Manual_2023-10_Rev-A.fodt

blattms: Done.

hakonhagland: Thanks for the update! When I CTRL-click the "github" link, nothing happens. Should it open in the default browser on my computer, or is this only the case after we have exported to PDF?

image

hakonhagland: > or is this only the case after we have exported to PDF?

Checked after exporting to PDF, and it works when I click the link in the PDF file.

PR #157, opened 2024-02-19
Update RPTRST.fodt

pkirkham:
Correct typo.


blattms:
Thanks  a lot for this fix. Works like a charm.

PR #160, opened 2024-02-24
Script to split git commit into two parts to make review of PRs easier

hakonhagland:
Builds on #159 which should be merged first.

Added a script `fodt-split-commit` that can be used to split a commit for a single `.fodt` file into two commits. The purpose is to make the review of a PR easier by splitting (usually) irrelevant changes in the meta section into a separate commit. Then reviewer can focus on the changes in the body part of the document. The requirement for using the script is that the last commit contain changes to a single .fodt file. The commit is then replaced by two commits.  The first one representing changes to the meta section of the `.fodt` file, and the second commit representing changes to the `office:body` section (the actual content of the file) of the `.fodt` file.

This script may not be extremely useful currently, as it cannot separate out style changes in the body part of the document yet. But it can work as a building block if we decide to make a more advanced script in the future.
    
Also added some test cases that can work as a starting point if we want to add more tests in the future.


scripts/python/tests/test_split_git_commit.py
```diff
@@ -0,0 +1,59 @@
+from pathlib import Path
+from typing import Callable
+
+from git import Repo
+# from pytest_mock.plugin import MockerFixture
```
lisajulia, line 5: Was this commented out line intentional?

scripts/python/src/fodt/split_git_commit.py
```diff
@@ -0,0 +1,242 @@
...
+# fodt-split-commit
+#
+# DESCRIPTION
+# -----------
+# Split the last committed file in a Git repository into separate commits for style and
+#  content changes. To simplify complexities with stashing of files, the last commit
+#  should contain changes to a single .fodt file. If the last commit contains
+#  changes to multiple files en error will be raised. The user is then required
+#  to manually separate the changes into separate commits before running this script.
+#  If the last commit contain change to a file that is not a .fodt file, an error
```
lisajulia, line 221: contain*s* change*s*

lisajulia (changes requested):
I've tried this splitting, it works nicely :+1: :smiley: !
One request though: The place to document all fodt-* commands is the scripts/python/README.md file, right? If yes, can you add the new *fodt-split-commit* command there?

scripts/python/tests/test_split_git_commit.py
hakonhagland: Yes, the test script does not currently use [pytest-mock](https://pypi.org/project/pytest-mock/) but will likely do when more tests are added in the future.

scripts/python/src/fodt/split_git_commit.py
hakonhagland: Thanks! I have updated the PR to fix the spelling

lisajulia (approved):

PR #161, opened 2024-02-25
Added GitHub actions workflow

hakonhagland:
Note that this PR builds on #160, which should be merged first.
 
Added GitHub actions workflow for testing Python scripts when a PR is submitted or modified. Currently it only runs tests in `scripts/python` but in the future it could be extended to run the tests in the `docker` directory also. The tests are only run if a file in `scripts/python` is changed or the github action workflow itself is modified.


blattms:
@akva2, do you have experience with github actions and are able to review this. I am not too confident on my own abilities in this regard.

blattms:
thanks, noob here, too.

PR #169, opened 2024-02-27
Fixes a bug in the Appendix heading numbering

### LINKED ISSUE #165: Appendix F is titled Appendix A
How best to change this as I believe it is part of main.fodt?

hakonhagland:
Builds on #168, which should be merged first.

Appendix E's heading tag had a missing property `text:continue-numbering="true"` which made the numbering for the next appendix to restart on A instead of continuing on F. 

Fixes #165


blattms:
Nice. thanks a lot.

PR #171, opened 2024-02-28
Created script to cleanup XML files

hakonhagland:
Note: this PR builds on #170 which should be merged first.

 Created a script that runs the Python `xml.sax` parser on all XML files and then writes back the content in the same way as LibreOffice writes XML. That is: it escapes `&`, `<`, `>`, `'`, and `"` and uses the short form `<tag/>` for empty tags instead of the long form  `<tag></tag>`.

By running this filter once on all files, we should hopefully get smaller PR diffs in the future since the Python scripts and LibreOffice should now agree on the format.

### DIFF
--- a/scripts/python/pyproject.toml
+++ b/scripts/python/pyproject.toml
@@ -35,6 +35,7 @@ fodt-split-main = "fodt.split_main:split_main"
 fodt-split-subdocument = "fodt.split_subdocument:split_subdocument"
 fodt-splitter = "fodt.splitter:main"
 fodt-validate-document = "fodt.validate_automatic_styles:validate"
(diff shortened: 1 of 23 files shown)


blattms (approved):
for my eyes this looks good. WOuld be good if @lisajulia would have another look and merge if she agrees.

hakonhagland:
> should we run this script once on all files and commit that?

Yes that was the idea! I have already submitted a PR that does that, see #172

lisajulia:
> > should we run this script once on all files and commit that?
> 
> @lisajulia Yes that was the idea! I have already submitted a PR that does that, see #172

Ahhh ok, sorry I had not seen the other PR ...

PR #172, opened 2024-02-28
Cleanup all XML files

hakonhagland:
Note: Builds on #171 which should be merged first.

Ran the cleanup script in #171 on all XML files. Hopefully this help getting smaller PR diffs in the future since the Python scripts and LibreOffice should from now on write the files the same way.


pkirkham:
Regarding the size of the PR diffs, what I've been doing is editing the XML in the .fodt file directly. It's confusing at first because the files contain so much additional stuff and so little actual content, but once you know where to look (near the end of the document) it's relatively quick and easy to find what needs to be edited.

However, I've only been doing small edits like typos, corrections to units etc. For a large edit you probably would need to open in LibreOffice.

In any event, I look forward to seeing cleaner XML in the documents. Does the script also get rid of all the text span tags which someone was saying are related to redlining? They are totally unnecessary and make it harder to work out exactly what needs editing.

hakonhagland:
> Does the script also get rid of all the text span tags which someone was saying are related to redlining? 

@pkirkham After this PR is merged the plan is to create another PR with a script to remove the span tags and then a PR where that script is applied to the XML files. The script will likely change many XML files and the current PR will help make the review of those changes simpler (since the changes will not also contain the formatting changes that the current PR includes)

hakonhagland:
Rebased

hakonhagland:
> We seem to have more than 1400 files

@blattms Yes :grin:

PR #173, opened 2024-02-29
Created script to remove revision span tags

hakonhagland:
LibreOffice uses span tags to track revisions, but we do not need this information since we are using Git for revision control. This script will try to remove those span tags and the corresponding style elements from the XML documents.

The plan is to apply this script to all XML documents to remove the revision span tags. By removing these tags, PR authors should have less trouble to do manual edits of the XML documents, and also the PR review should become easier.


lisajulia (commented):
All is fine

PR #176, opened 2024-03-01
GASDENT add description of correlation

gdfldm:
Also remove sentence "The reference pressure for this table is given by the VISCREF keyword in the PROPS section." as I don't think it is relevant for this keyword.


blattms:
Looks great and much more useful now.

PR #180, opened 2024-03-04
Address various comments (part 1)

gdfldm:


blattms:
Easy review thanks to xml-editing. In it goes.

PR #201, opened 2024-03-12
Update H2STORE

gdfldm:
Reordered options 1 and 2 to put GASWAT first
Move DISGASW before VAPWAT and PRECSALT
Reordered option 1 and 2 examples


blattms (commented):
Looks good except for one typo (might be old, though). Please accept my suggestion if you agree and I'll merge

PR #203, opened 2024-03-13
 Correct numerous typos and inconsistencies in manual (part 2)

blattms:
A second set of changes that I've been accumulating whilst going through the manual.

This is a rebased version of #202 of @pkirkham 

Closes #202

PR #204, opened 2024-03-13
Updated README.md with information about disabling document comparison option

hakonhagland:
See https://github.com/OPM/opm-reference-manual/pull/203#discussion_r1523537605 for background.

Added information about disabling document comparison option in LibreOffice to make editing documents easier.


blattms:
This is of course  a global setting in libreoffice and people might not want to have this elsewhere.

hakonhagland:
@blattms Yes I did not think about that. Maybe we could make the docker version use this setting by default? Since the docker version of libreoffice will only be used for editing the opm-reference-manual files and the docker container will have its own settings (separate from the host machine settings)

blattms:
I added a suggested change. With that I would merge this anyway.

hakonhagland:
@blattms Thanks! Updated.

blattms:
Cool. Thanks a lot.

PR #206, opened 2024-03-13
Add script to remove undefined span tags

hakonhagland:
For background, see https://github.com/OPM/opm-reference-manual/pull/203#discussion_r1523590062

Added script to remove undefined span tags from fodt files. A span tag is considered undefined if it refers to a style that is not defined in the document.


lisajulia (approved):
tested and looks good to me

PR #207, opened 2024-03-13
Removes undefined span tags from all fodt files

hakonhagland:
Builds on #206, which should be merged first.

For background see #206 and #203. Applied the script in #206 to remove undefined span tags.


gdfldm:
@hakonhagland I have resolved the conflicts, these were just typos that had been fixed elsewhere

Issue #67, opened 2024-01-06
How to get rid of the popup dialog "Would you like to update all links to get the most recent data?"


hakonhagland:
Each time I open the main document or a chapter document with libreoffice a popup dialog appears with the message

 _"The document contains one or more links to external data. Would you like to change the document, and update all links to get the most recent data?"_

How to automatically select “Yes” to this question and avoid having to answer this dialog each time?

PR #208, opened 2024-03-14
Copy custom config files to docker image

hakonhagland:
Copies a snapshot of LibreOffice config files into the docker image. The config files have been modified from the default LibreOffice config files in two ways:

- First, document comparison mode has been disabled, see PR #204 for details,
- Second, the parts sub directory inside the image has been added as a trusted location to get rid of the popup dialog "Would you like to change the document, and update all links to get the most recent data?", see https://github.com/OPM/opm-reference-manual/issues/67 for details.


hakonhagland:
@blattms I am not sure why we we would use libreoffice to compare two different documents when we already have git version control? Can you elaborate on that idea?

Anyway, as I remember it, this pull request was in accordance with the README.md file which suggests that we turn off document comparison mode. If the user still want document comparison he could he could avoid using the docker container version of libreoffice or he could manually re-enable the mode from the menu in libreoffice. But it is not recommended if he plans to submit a clean PR..

blattms:
> @blattms I am not sure why we we would use libreoffice to compare two different documents when we already have git version control? Can you elaborate on that idea?

Because humans cannot read formatting changes. It is really hard to compare changes using git only. I'll check that myself once I find time.

hakonhagland:
> Because humans cannot read formatting changes. It is really hard to compare changes using git only. I'll check that myself once I find time.

@blattms Can you check then if it would be a possible strategy for a user to have document comparison mode enabled and only disable when submitting changes to the git repository? Then we can avoid tracking the revision ids in the xml files which should simplify the git diff review process according to the dicussion about span tags in #203.

blattms:
I will if you tell me how to do this,

blattms:
This is what I am using currently for quick comparison (use "File -> Track Changes -> Compare Documents" and use a copy of the old version):
image

hakonhagland:
> This is what I am using currently for quick comparison (use "File -> Track Changes -> Compare Documents" and use a copy of the old version):

@blattms Thanks for that. What is the problem with that workflow that makes you want to keep the revision ids in the XML files? Sorry, it is not clear to me yet why this workflow is in conflict with this PR.

blattms:
Can you please remind me how to use the docker image. I am trying this for while now and always fail (e.g. the file `/home/dockeruser/parts/appendices/A.fodt is not found. My approach

- go to src python, create venv and `pip install .`
- got to docker and do `pip install .`
- in the same directory run `lodocker-run-container parts/appendices/A.fodt`

hakonhagland:
> the file `/home/dockeruser/parts/appendices/A.fodt is not found

@blattms Added a fix to the script here: https://github.com/OPM/opm-reference-manual/pull/255. Please check if that works.

PR #211, opened 2024-03-15
Final 2023-10 version for distribution

blattms:
this updates all the document controls and document contributors. With this we should be able to release this. Good if @gdfldm takes another look.


gdfldm (approved):
Thanks, these changes look good to me.

blattms:
Finally published. See https://opm-project.org/?p=1645

PR #217, opened 2024-03-18
Corrected wrong letter "E" to letter "K" in footer in all sub documents

hakonhagland:
Note: this PR builds on #216 which should be merged first.

Applied script from #216 to all documents.


blattms:
Didn't we fix the same thing before? How did it break again?

hakonhagland:
> Didn't we fix the same thing before? How did it break again?

@blattms Seems like we only fixed it for main.fodt, see https://github.com/OPM/opm-reference-manual/pull/68

PR #218, opened 2024-03-25
Typos and minor edits #3

pkirkham:
I believe I've correctly re-baselined this time before making a bunch of changes. Hopefully the underlying documents haven't changed too much in the past few days since I started marking up the changes.


gdfldm (approved):
@pkirkham thank you for edits.

blattms:
@gdfldm If you have tested and approved a contribution then it is also fine to merge it. Waiting for me might result in rather long delays :sweat_smile: and my approval is actually  less worth than yours.

Issue #219, opened 2024-04-04
ROCKTAB


hnil:
The description say that this is unsupported, however the code use it.

gdfldm:
Thanks for spotting this. The manual has been updated, see #237.

PR #220, opened 2024-04-04
Mark ROCKTAB as supported

gdfldm:
in keyword description and appendix a
Addresses #219


blattms:
I guess @gdfldm  has used our scripts for this.  @hakonhagland, might the add_keyword_status.py mess up the diff due changing xml-end-tags?

hakonhagland:
> might the add_keyword_status.py mess up the diff due changing xml-end-tags?

@blattms Yes you are right, it would. I think the end tags was fixed in https://github.com/OPM/opm-reference-manual/pull/227 but that is not merged yet.

gdfldm:
> who did you perform these changes? With scripts?

@blattms I used the script to update Appendix A. Then edited the XML for ROCKTAB.fodt.

gdfldm:
replaced by #237

PR #221, opened 2024-04-04
Mark ACTION, ACTIONG and ACTIONR as unsupported in Appendix A

gdfldm:


blattms:
Without having looked at this in any detail: Why is ACTIONX unsupported?

gdfldm:
> Without having looked at this in any detail: Why is ACTIONX unsupported?

@blattms oops, I have updated the title.

gdfldm:
Replaced by #238

PR #223, opened 2024-04-04
Delete duplicate table row for AQANNC in Appendix A

hakonhagland:
Delete duplicate table row for AQANNC in Appendix A. Thanks to @gdfldm for spotting.

@gdfldm Can you test if this looks OK from main.fodt?


hakonhagland:
> I just noticed in the XML for the duplicate entry there is a reference to AQANCONL, which doesn't appear in Appendix A. I think this may have been a typo and the first entry should probably be AQANCONL.

@gdfldm Yes you are right, I created a new PR to replace this. See #224.

PR #225, opened 2024-04-04
Fixed typo in GASDENT in Appendix A

hakonhagland:
Thanks to @gdfldm for spotting this.


gdfldm:
@hakonhagland thanks for the fix. That was my first merge, hope it worked ok :)

Issue #228, opened 2024-04-08
add-keyword script creates messy diffs


blattms:
If I do (my intention was to show how clean the diff of the other files is :sweat_smile: ):
```
$ git reset --hard HEAD
$ fodt-add-keyword --maindir=../parts --keyword=HELLO --section=4.3 --title="Hello World" --status=green
````
the diff looks a bit messy. Seems like the script changes A.fdot by reintroducing closing tags, e.g.:
```diff
-   <config:config-item config:name="PrinterSetup" config:type="base64Binary"/>
+   <config:config-item config:name="PrinterSetup" config:type="base64Binary"></config:config-item>
```

Not sure where this comes from. I thought we actually changed that.

hakonhagland:
Yes I think we have not fixed the add-keyword script to handle empty tags yet. I will try to fix the script.

hakonhagland:
Closing as this was addressed in #229.

PR #212, opened 2024-03-15
Added a more advanced method to locate the maindir from the Python scripts

hakonhagland:
For background see #209.

Added a more advanced method to locate the `maindir` and `filename` arguments for usage with the Python scripts. This is useful because the Python scripts can be installed globally and then run from any directory. The original logic assumed that the Python scripts were run from `scripts/python`. Also added some unit tests.

PR #227, opened 2024-04-05
Update the set-keyword-status script

hakonhagland:
Updated the `fodt-set-keyword-status` script to also handle the case when keyword is specific to OPM flow:

image

when the keyword is specific to OPM flow, the text "OPM Flow" is shown in the table as shown in the screenshot above. The script can now be run like this:

```
$  fodt-set-ketword-status --keyword=CSKIN --color=green --opm-flow
```
to put the text "OPM Flow" or just omit the flag `--opm-flow` to remove the text "OPM Flow".

Thanks to @gdfldm for the suggestion.

@gdfldm Can you test this?

PR #229, opened 2024-04-08
Fix add keyword

hakonhagland:
This PR builds on #212, which should be merged first.

For background see #228. Updated the add-keyword script to handle empty tags.


blattms:
Does this also automatically apply to add_keyword_status.py?

hakonhagland:
> Does this also automatically apply to add_keyword_status.py?

@blattms No, but the set-keyword-status script was fixed to handle empty tags in #227 I believe

hakonhagland:
@lisajulia Sure!

lisajulia (approved):
Looks good to me; I've tested this as well.

PR #231, opened 2024-04-09
Cleaned up all xml meta files

hakonhagland:
This PR builds on #230, which should be merged first.

Used the script in #230 to clean up xml meta files.


gdfldm:
@hakonhagland, I have spot checked the changes to the XML and created a PDF from main.fodt, and it looks good. Thank you!

PR #235, opened 2024-05-14
Add ACTCO2S to PROPS section

svenn-t:
This PR adds description of the ACTCO2S keyword to the PROPS section. The keyword is used to choose between different salting-out effects when computing mutual solubility between CO2 and brine in the CO2STORE module.


gdfldm:
Superseded by reformatted version #246

PR #239, opened 2024-05-20
Fixed bug in fodt-xml-sax-filter-meta script

hakonhagland:
A bug was introduced in #230 and #231 such that the xml header tag was added to all meta files.

This PR removes the headers from the meta files (introduced in #231) and fixes the `fodt-xml-sax-filter-meta` script to work correctly (not adding the header to the meta files)

Thanks to @gdfldm for spotting the problem.


gdfldm:
@hakonhagland. Thank you for the fix.

PR #245, opened 2024-05-23
Update COMPDAT to support DFACT for 2024-04

gdfldm:


blattms:
Looks good.

Note to self: Document comparison in libreoffice does not work well for tables. Seem like the whole table is marked as changed (if some text changes).

blattms:
Yep adding one character to a table triggers this behavior (Table remove and new Table added)

PR #238, opened 2024-05-20
Mark ACTION, ACTIONG and ACTIONR as unsupported V2

gdfldm:
fodt-set-keyword-status --keyword ACTION --color orange 
fodt-set-keyword-status --keyword ACTIONG --color orange 
fodt-set-keyword-status --keyword ACTIONR --color orange

PR #248, opened 2024-05-23
Fixed a bug in the fodt-set-keyword-status script

hakonhagland:
For background see https://github.com/OPM/opm-reference-manual/pull/238#discussion_r1611595172.

A newline was accidentally gobbled up by the `fodt-set-keyword-status` script. Hopefully this change fixes the issue.


hakonhagland:
Rebased this.

gdfldm:
Thank you for the fix.

PR #249, opened 2024-05-23
Add ACTCO2S keyword description

gdfldm:


blattms:
Can you explain this please? File was there and is somehow changed now. What is changed, was there some fallout?

gdfldm:
> Can you explain this please? File was there and is somehow changed now. What is changed, was there some fallout?

The rebased PR #247 added the keyword fine. But only a blank keyword template was added without the description, so I have added again here.

blattms:
then that was my fault. I am sorry about that,

PR #250, opened 2024-05-23
Mark DATUMR as supported for 2024-04

gdfldm:


blattms:
Looks good.

Looking at the diff I doubt that the color changes in this PR. It is green though. Maybe I missed something.

blattms:
My impression was wrong. The style was changed. I just got confused as in the actual file the color is changed directly by our script.

PR #254, opened 2024-05-28
Update ACF keyword description

gdfldm:


blattms:
Thanks  a lot for the update and corrections

PR #255, opened 2024-05-28
Fixed bug in the run docker container script

hakonhagland:
For background see https://github.com/OPM/opm-reference-manual/pull/208#issuecomment-2134838443.

The git root will not always be 3 levels up from the filename. Added a more advanced method to find the git root. Also added a test case.


hakonhagland:
>Seemed like it had to be created below scripts/python

@blattms Yes, I tried now with "pip install ./docker" from the root directory, and there seems to be some assumptions in the script about the current directory being "<root>/docker" when running the docker python scripts. I will check what is going on.

lisajulia:
Tested and it works :)

PR #259, opened 2024-05-28
Update add keyword script with new option

hakonhagland:
Thanks to @gdfldm for the suggestion.

Added a new option `--no-appendix` to the `fodt-add-keyword` script. This can be used to avoid adding the keyword to the appendix in the rare case that you want to add two keywords with the same name to different sections.


gdfldm:
@hakonhagland. Thank you for the update.

PR #261, opened 2024-05-28
Add EOS keyword to RUNSPEC and PROPS sections as unsupported

gdfldm:


blattms:
Should we close this as #262 is merged?

gdfldm:
Yes, closing

PR #263, opened 2024-05-29
Add EOSNUM keyword to REGIONS section as unsupported

gdfldm:


gdfldm:
@blattms. I have replaced this with #277.

PR #275, opened 2024-06-04
Update SKIP300 keyword for 2024-04

gdfldm:


gdfldm:
> Is this correct, that this only skps "compositional" keywords? Sounds strange. Additionally we always talking about a "commercial" simulator here.

@blattms The behaviour is different for the commercial b-o and compositional simulators. Flow 2024-04 is different again. But a command line option will be added for future versions of Flow to give the option of b-o or compositional like behaviour.

PR #277, opened 2024-06-05
Add EOSNUM keyword to REGIONS section as unsupported (manually rebased)

gdfldm:


gdfldm:
@blattms this has been replaced by #317

PR #282, opened 2024-06-06
Add SOURCE keyword to SCHEDULE section for 2024-04

gdfldm:
fodt-add-keyword --keyword SOURCE --section 12.3 --title "Define Source Term" --status green --appendix

PR #283, opened 2024-06-06
Fix SGU link in appendix A

hakonhagland:
For background see https://github.com/OPM/opm-reference-manual/pull/282#issuecomment-2152581962

Fixes the link to SGU from Appendix A to Chapter 8.


gdfldm:
Thank you for the fix

PR #286, opened 2024-06-06
Add SOURCE keyword to SCHEDULE section for 2024-04 v3

gdfldm:
Includes modification to regex in add-keyword status script (as in #285).

Closes #282


totto82:
Thanks for the contribution. The text looks good to me.

PR #291, opened 2024-06-11
Minor edits #4

pkirkham:
Corrected minor typos in new supported keywords for 2024.04. Added missing terminating slashes for sub-table records of second table in examples for PVTG series of keywords.


blattms:
Would you mind reviewing @gdfldm

gdfldm (approved):
Thanks for the edits. These look good to me.
We may also want to add an extra row (sub-table) at the end of table no.2 in the examples for PVTGW and PVTGWO. But that can be a new PR.

PR #300, opened 2024-06-12
Update OPERATE

gdfldm:


blattms:
I am under the impression that some arrays won't work with `OPERATE`. Do happen to know whether there are tests in combination with `MULTZ`/`TRANZ`?

gdfldm:
> I am under the impression that some arrays won't work with `OPERATE`. Do happen to know whether there are tests in combination with `MULTZ`/`TRANZ`?

I found a couple here...

[MULTZ](https://github.com/OPM/opm-tests/blob/9891c043164ed40e2d572a0cb48b43594a98d4a2/operate/OPERATE-MULTZ.DATA)
[TRANZ](https://github.com/OPM/opm-tests/blob/9891c043164ed40e2d572a0cb48b43594a98d4a2/operate/OPERATE-TRANZ.DATA)

PR #320, opened 2024-07-08
Repair DISPERSE keyword

gdfldm:
Links to the DISPERSE keyword description seemed to be broken. This seems to be due to there being no space before the dash in the title of the page i.e. "DISPERSE- etc" not "DISPERSE - etc".


blattms:
I don't really get why the behavior is different, but I trust you.

PR #322, opened 2024-07-08
Update Appendix E

gdfldm:
Insert section E.1 running OPM Flow 2023-10


blattms:
- Option --ecl-newton-sum-tolerance-exponent: double the in  "The the exponent used to scale ..."
- Option -enable-grid-adaptation: maybe mention that this will switch off eclipse output
- Option no. 53/54  should probaly be moved to another category. These are not related to ILU0
- Option no. 64-70, 74, 101-103: these seem to be related to Nonlinear Local Domain Decomposition (NLDD) as the linear solver. Should we put those in a separate category. On the other hand anything fro 62 on is not about the linear solver anymore. So maybe just start  a misc category?
- Option no. 84 is about ILU0 so maybe move?
-  Option 105-106 is about linear solver. Maybe move?
- the rest until the next category is somehow in "Misc" and not the current category
- Option 165 is about the linear solver

gdfldm:
@blattms I wonder if we could remove the category headings? The options are currently listed in alphabetical order.

blattms:
Whatever is better for your workflow and saves time.

gdfldm:
Thanks, I'll remove the category headings :)

PR #323, opened 2024-07-08
Update Section 2 Running Flow 2024-04 etc

gdfldm:


blattms:
There is a bug in our packages (libopm-simulators-bin is missing python3-opm-common as a dependency. The latter is needed if someone uses PYACTION.
Hence please replace `sudo apt-get install libopm-simulators-bin` with `sudo apt-get install libopm-simulators-bin python3-opm-common`
To be on the safe side we should probably do the same for Redhat: `sudo yum install opm-simulators-bin python3-opm-common`

atgeirr:
I assume this means we should also update the installation instructions here: https://opm-project.org/?page_id=245, can you take care of that as well?

PR #327, opened 2024-07-09
Mark GASVISCT as Supported in Appendix A

gdfldm:
fodt-set-keyword-status --keyword GASVISCT --color green


gdfldm:
> This removed the "OPM FLOW" specific marker. Is that intended?

Oops!

PR #333, opened 2024-07-17
Repair links to AQANCONL

gdfldm:


blattms:
I am starting to wonder whether libreoffice just changes `text:ref-name=` whenevever it likes, Then we hve a rpoblem...

gdfldm:
I hope not that would be a nightmare!

Issue #54, opened 2024-01-02
cross-references between subdocuments are broken


blattms:
E.g. in ./parts/chapters/subsections/12.3/GRUPNET.fodt there is  a reference to figure 12.4 in ./parts/chapters/subsections/12.3/GRUPTREE.fodt which is unkown "Error: Reference not found"

gdfldm:
It looks like reference names are not unique, refTable6 appears in 5 different fodt files, for example:
```
<text:sequence text:ref-name="refTable6" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">2.7</text:sequence>
```
```
<text:sequence text:ref-name="refTable6" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">11.2.7</text:sequence>
```
This might make it difficult to reference a table from a different sub-document file :(

blattms:
Not sure whether we can resort to bookmarks. See [this issue](https://superuser.com/questions/1594796/cross-referencing-from-a-master-document-error-reference-source-not-found)

blattms:
Hopefully these reference names are persistent. In that case we can come up with a scheme for their name (e.g. section_5_3__table_1 etc.) and rename them. See this [question](https://superuser.com/questions/1235922/how-to-rename-a-reference)

PR #334, opened 2024-07-17
Repair reference to equation in WSEGAICD

gdfldm:


blattms:
This is #54. The reference names might get renumbered within a document without the other documents recognizing it and then it does not fit any more. This will happen whenever we open one document with libreoffice and add new stuff.

PR #335, opened 2024-07-17
Remove references to tables in MESSAGES

gdfldm:
The table references are unnecessary as there is a link to the page containing the table.

### DIFF
--- a/parts/chapters/subsections/4.3/MESSAGES.fodt
+++ b/parts/chapters/subsections/4.3/MESSAGES.fodt
@@ -4262,7 +4262,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
      </table:table-row>
     </table:table>
     <text:h text:style-name="Heading_20_4" text:outline-level="4" text:is-list-header="true"><text:bookmark-start text:name="__RefHeading___Toc32167_370116838818"/>Description<text:bookmark-end text:name="__RefHeading___Toc32167_370116838818"/></text:h>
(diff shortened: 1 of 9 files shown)


parts/chapters/subsections/4.3/MESSAGES.fodt
```diff
@@ -4506,7 +4506,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:covered-table-cell/>
      </table:table-row>
     </table:table>
-    <text:p text:style-name="P14560">Table <text:sequence text:ref-name="refTable15" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">4.5</text:sequence>: MESSAGES Keyword Description</text:p>
+    <text:p text:style-name="P14560">Table <text:sequence text:ref-name="REF_TABLE_MESSAGES_KEYWORD__CRIPTION_4_3" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">4.5</text:sequence>: MESSAGES Keyword Description</text:p>
```
blattms, line 4509: Did you do the renaming of text:ref-name?

blattms:
But they also don't really hurt. Why do you think we should remove them?

parts/chapters/subsections/4.3/MESSAGES.fodt
gdfldm: Yes I tried to create a unique ref-name for the table number. But it didn't work.

gdfldm:
> Why do you think we should remove them?

Because I couldn't find anyway to have a working table reference between sub-documents. (I works fine within the same sub-document). 

The only way that I can think of that might work is to wrap the table caption in a bookmark or a cross-reference.

PR #336, opened 2024-07-18
Fix table references to SATOPTS in KRNUM and IMBNUM

gdfldm:
~Change refs to tables to refs to keyword~

Table refs have now been fixed:

I have now wrapped the Table number 'Table 5.40' in SATOPTS in a reference and tried to give it a unique ref-name. This ref-name is then cross-referenced from IMBNUM and KRNUM. I have successfully tested the links in main.fodt.


blattms:
Why is that better?
Do the ref-names of tables get changed automatically? If not we could use a more unique scheme and rename them. See links in #54

PR #339, opened 2024-07-19
Fix Table Refs in Section 2 4

gdfldm:
~I was unable to get table references between sub-documents to work, so I have used a reference to the section instead.~

Fixed table references in section 2.4

### DIFF
--- a/parts/chapters/sections/8/2.fodt
+++ b/parts/chapters/sections/8/2.fodt
@@ -19612,7 +19612,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       </table:table-cell>
      </table:table-row>
     </table:table>
-    <text:p text:style-name="P77">Table <text:sequence text:ref-name="refTable5" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">8.6</text:sequence>: Oil-Water Relative Permeability End-Point Data Definitions</text:p>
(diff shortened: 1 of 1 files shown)


blattms:
I think I messed this one up with one of my merges.
It needs a rebase, sorry.

parts/chapters/2.fodt
```diff
@@ -78973,9 +78973,9 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
-    <text:p text:style-name="P449">Finally, see the SCALCERS keyword in the PROPS section, that defines the options used in the re-scaling process, including the options for two point and three point scaling. Whatever option is selected, always verify that the de-normalized curves that will be used in the model are correct and reasonable.</text:p>
-    <text:p text:style-name="P446"/>
+    <text:p text:style-name="P447">As mentioned previously, entering a correct original curve that is not normalized will automatically resolve these type of issues. If a normalized curve is entered, then ensure that simulator can distinguish between SWL and SWCR etc., based on the variable definitions provided in <text:reference-ref text:reference-format="text" text:ref-name="REF_TABLE_OIL_WATER_RELATI__INITIONS_8_2">Table 8.6</text:reference-ref>,<text:s/><text:reference-ref text:reference-format="text" text:ref-name="REF_TABLE_GAS_OIL_RELATIVE__INITIONS_8_2">Table 8.7</text:reference-ref><text:s/>and<text:s/><text:reference-ref text:reference-format="text" text:ref-name="REF_TABLE_GAS_WATER_RELATI__INITIONS_8_2">Table 8.8</text:reference-ref>. </text:p>
```
blattms, line 78976: On my system there are a lot of page breaks in this area. Do you see that, too?

blattms (commented):
References work.
In 2.fodt there seem to be unmotivated page breaks. These were not there before and I don't see how they got introduced.

parts/chapters/2.fodt
gdfldm: Yes, looks like I broke the paragraph formatting during the rebase. The new commit should fix this.

PR #345, opened 2024-07-24
Fix figure reference in GRUPNET

gdfldm:
Replaces PR #344

I have wrapped the Figure number 'Figure 12.4' in GRUPTREE in a reference and tried to give it a unique ref-name. This ref-name is then cross-referenced from GRUPNET. I have successfully tested the links in main.fodt and in the exported main.PDF.


blattms:
Very cool. Thanks.

PR #359, opened 2024-08-01
Correct keyword name MISNUM -> MISCNUM

gdfldm:


blattms (changes requested):
Consequently we should also change the filename to MISCNUM.fodt and correct keywords.txt accordingly.

blattms (approved):
Which is already done. Sorry for the noise.

PR #360, opened 2024-08-01
Text updates Section 10.1

gdfldm:


parts/chapters/10.fodt
```diff
@@ -19214,18 +19214,18 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+    <text:p text:style-name="P12284">The SOLUTION section defines the initial conditions for the model, that is the conditions at time zero, before any flow occurs in the reservoir. <text:s/>As such, sufficient information from the previous sections is required for the simulator to calculate the initial saturation and pressure distributions through out the model, such that the fluids are in gravity/hydrostatic equilibrium. This means that if the simulator is run for a period of time without any production or injection then there should be no saturation or pressure changes in the model. <text:s/>In fact this “test” of running the simulator for a few years without any production or injection should always be performed to ensure that the model is indeed in hydrostatic equilibrium.</text:p>
```
blattms, line 19217: Is the additional space <text:s/>  before As intended?

gdfldm: LO seemed to me to make the space after the full stop too small without the additional space. But happy with or without it :)

gdfldm:
@markus, with equation (10.2) etc, I wasn't sure if it was best to redefine gamma as a pressure gradient or add a gravity term to the equation? (But it was easier to edit the text than the equations in XML.)

PR #383, opened 2024-08-22
Misc small changes/errata

pkirkham:
A couple of minor changes spotted in the SATOPTS and SGWFN keywords. Added note concerning PERMX, PERMY and PERMZ grid cell variables with OPERATE keyword which is computed incorrectly if using FIELD or LAB system of units. See: https://github.com/OPM/opm-simulators/issues/4597


gdfldm:
Thank you for spotting these. 
Since permeability is input in mD for all units systems, doesn't this issue also apply when using METRIC units?
How about dividing by 1.01325e-15 instead?

pkirkham:
I'd have expected units of mD to be treated identically for all systems of units, but that isn't the case. I hadn't looked too closely at the conversion factor (I just used what was mentioned in issue #4597). Upon further consideration, dividing by 1.01325e-15 is more accurate. I seem to recall that I tested with METRIC, FIELD and LAB, and found that the issue only exists for FIELD and LAB (which is weird). If I was more adept at C/C++ I'd try to dive into the code and find where the problem is. However, there is something of a learning curve to go up in understanding exactly where the problem is located, so I thought in the meantime just updating the manual would be a reasonable stop-gap measure.

I'll try to update this pull request.

gdfldm:
I agree, the manual should reflect the current version of the simulator even if its behaviour is a bit weird.

blattms:
I'l leave merging to @gdfldm

PR #399, opened 2024-10-09
Mark items that support UDAs: WCONINJE, WCONPROD

gdfldm:
Based on definitions in
opm/input/eclipse/share/keywords/*


blattms:
For diffs like this git makes a better job comparing versions. With libreoffice`s "Compare document" any change in a table is shown as new table added and old remove

PR #402, opened 2024-10-10
MULTREGD and MULTREGH add note on defaulted region numbers

gdfldm:
Depends on #401

### DIFF
--- a/parts/chapters/subsections/6.3/MULTREGH.fodt
+++ b/parts/chapters/subsections/6.3/MULTREGH.fodt
@@ -4295,7 +4295,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P3659">REGION1</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table1397.C7" office:value-type="string">
-       <text:p text:style-name="P3659">A positive integer value that defines the from REGION number for which the CONSTANT in (2) should be applied.</text:p>
+       <text:p text:style-name="P3659">A positive integer value that defines the from REGION number for which the CONSTANT in (3) should be applied.</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table1397.D7" office:value-type="string">
        <text:p text:style-name="P3250">None</text:p>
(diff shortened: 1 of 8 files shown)


parts/chapters/subsections/6.3/MULTREGH.fodt
```diff
@@ -4400,6 +4401,15 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
         <text:list-item text:start-value="1">
          <text:p text:style-name="P19105">Where REGION1 and REGION2 should be less than or equal to the maximum number of regions as defined on the REGDIMS keyword for the FIPNUM and OPERNUM arrays or the GRIDOPTS keyword for the MULTNUM array in the RUNSPEC section.</text:p>
         </text:list-item>
+        <text:list-item>
+         <text:p text:style-name="P19105">If REGION1 is defaulted then the transmissibilities between REGION2 and all other regions will be multiplied. Similarly, if REGION2 is defaulted then the transmissibilities between REGION1 and all other regions will be multiplied. If both REGION1 and REGION2 are defaulted then the transmissibilities between all pairs of regions will be multiplied.</text:p>
+        </text:list-item>
+        <text:list-item>
+         <text:p text:style-name="P19105">If REGION1 and REGION2 are equal and positive then all the transmissibilities within that region as well as the transmissibilities between that region and all other regions will be multiplied.</text:p>
```
blattms, line 4408: We might want to add another here and elsewhere (MULTREG*) in a separate PR. There is another specialty if the regions are the same. If a multiplier has been  specified using different regeions already then  a multiplier specified here will be accumulated to this multiplier (in all other cases, it would be overwritten...). See [MULTREGTScanner::applyMultiplierSameRegion](https://github.com/blattms/opm-common/blob/24a35e3e50ede23d49064e87c26bd41ce924428b/src/opm/input/eclipse/EclipseState/Grid/MULTREGTScanner.cpp#L446) and (bad) [comment](https://github.com/blattms/opm-common/blob/24a35e3e50ede23d49064e87c26bd41ce924428b/opm/input/eclipse/EclipseState/Grid/MULTREGTScanner.hpp#L152)

blattms (approved):

PR #403, opened 2024-10-10
Mark Items that support UDAs: WCONHIST, WPOLYMER, WECON, WSEGVALV, WSALT, WTMULT

gdfldm:


blattms (approved):
Thanks a lot looks good.

PR #408, opened 2024-11-04
Script to extract keyword URIs

hakonhagland:
This script can be used to extract a mapping from keyword names to XML URI tag names. The mapping can be saved to a file and later be used to make keywords clickable. So the goal is to enable the user to click on a keyword referenced in a paragraph of text in the manual, and then automatically jump to the page where that keyword is defined.

The plan is to use the produced mapping in another script (yet to be written) that will convert keyword names into `<text:a xlink:href="some_uri" ...>KEYWORD_NAME</text:a>` which will make the text KEYWORD_NAME into a clickable link.

### DIFF
--- a/scripts/python/src/fodt/keyword_linker.py
+++ b/scripts/python/src/fodt/keyword_linker.py
@@ -45,6 +45,8 @@ class ExtractURI_Handler(xml.sax.handler.ContentHandler):
         if name == "text:h":
             if self.in_section:
                 self.in_section = False
+                # The keyword URI must be found within the text:h, since we are done parsing
+                # the tag, we raise an exception here to catch an unexpected situation.
                 raise ParsingException("Keyword name not found in document")
         elif self.in_section and name == "text:bookmark-start":
             if self.in_bookmark:
--- a/scripts/python/src/fodt/keyword_linker.py
+++ b/scripts/python/src/fodt/keyword_linker.py
@@ -1,0 +1,161 @@
+import logging
+import re
+import xml.sax
+import xml.sax.handler
+import xml.sax.xmlreader
+import xml.sax.saxutils
+
+from pathlib import Path
+
+import click
+
+from fodt.constants import ClickOptions, Directories, FileNames, FileExtensions
+from fodt.exceptions import HandlerDoneException, ParsingException
+from fodt import helpers
+from fodt.xml_helpers import XMLHelper
+
+class ExtractURI_Handler(xml.sax.handler.ContentHandler):
+    def __init__(self, keyword_name: str) -> None:
+        self.keyword_name = keyword_name
+        self.in_section = False
+        self.in_bookmark = False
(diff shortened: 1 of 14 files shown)


scripts/python/src/fodt/keyword_linker.py
```diff
@@ -0,0 +1,161 @@
...
+        self.keyword_name = keyword_name
+        self.in_section = False
+        self.in_bookmark = False
+        self.in_bookmark2 = False
+        self.uri = ""
+
+    def characters(self, content: str):
+        if self.in_bookmark2:
+            # Check if the content matches the keyword name
+            match =  re.search(self.keyword_name, content)
```
blattms, line 28: I don't think that this enough. Aren't there keywords that include other keywords in their name? Hence we need an exact match. Maybe search for "^"+self.keyword_name+"$"`, or something?

scripts/python/src/fodt/keyword_linker.py
```diff
@@ -0,0 +1,161 @@
...
+            raise HandlerDoneException("Done parsing.")
+
+    def endDocument(self):
+        raise ParsingException("Keyword name not found in document")
+
+    def endElement(self, name: str):
+        if name == "text:h":
+            if self.in_section:
+                self.in_section = False
+                raise ParsingException("Keyword name not found in document")
```
blattms, line 48: Can you please comment and document why we raise an exception here? It is not immediately clear to me why the keyword name was not found when we throw ParsingException.

The name ParsingException seems a bit misleading. This might be personal taste, though.

blattms (commented):
Thanks a lot.

Would be cool, if there would be a bit more documentation.

The only thing I am wondering about is whether the regex matching will work as expected.

scripts/python/src/fodt/keyword_linker.py
blattms: Looking at the other PR, I seem to be wrong, though. Would you please enlighten me why/how?

hakonhagland: > I don't think that this enough.

@blattms Actually the regex should in theory not be necessary, but I kept it there as a control if something weird  happened. Since we already know that we are inside the file with the given keyword name and we are in the first  heading with outline level 3, it should be enough to extract the URI link from the first `<text:bookmark-start>` tag. Anyway, matching the keyword name exactly was also abondoned due to the other issue with span tags that might occur in the middle of the keyword and since it in theory should be necessary to check for it as mentioned above.

scripts/python/src/fodt/keyword_linker.py
hakonhagland: > why we raise an exception here?

It is just for documentation that this should never happen  (and also to catch an unexpected state early which will help with debugging) . If we reach this part of the code we will not have any hope of finding the URI value (since it should always be at the top of heading level 3 in the keyword file).

PR #409, opened 2024-11-04
Keyword to URI mapping file

hakonhagland:
Used the script in #408 to generate the mapping file. See #408 for more information.


parts/meta/kw_uri_map.txt
```diff
@@ -0,0 +1,1142 @@
+ACF REF_HEADING_KEYWORD_ACF
+ACTCO2S REF_HEADING_KEYWORD_ACTCO2S
+ACTDIMS __RefHeading___Toc4408_421927891
```
blattms, line 3: Do we trust libreoffice to not change uris ever?

hakonhagland: No I guess not. So this file should then be updated if we need to apply the scripts again. But I guess the scripts will only be applied once? Then maybe we should rather keep this file outside version control?

gdfldm: What if we later create a new keyword or update the text of existing keyword descriptions to mention new keywords - could the script be re-run to add the required new links, or would it be best to manually create the new links at this point?

hakonhagland: > could the script be re-run to add the required new links

Good point. I think it would be best to have a script that can be run. Either after each added keyword or delay running it until a new release of the manual is prepared. I think manual insertion after each edit may be to cumbersome?

I will check if we need to make changes to the script as it is now. I wrote it as a one-time script, if we plan to use it regularly we should add some documentation and maybe the script should automatically create this file (such that user only needs to run one script).

blattms:
So, we don't merge this?

hakonhagland:
> So, we don't merge this?

I have not seen yet that libreoffice changes the URIs in the file. But I guess it can happen, and then the file needs to be regenerated. I have no strong opinion on whether to merge this or not. If we don't merge it the script in #408 should probably not write the file to `parts/meta` as it does now? Or we could still write it there but add a `.gitignore` rule to avoid merging it.

PR #410, opened 2024-11-04
Script to insert keyword links

hakonhagland:
Builds on #408 and #409 which should be merged first.

The script replaces keyword names inside `<text:p ...>` tags with links to the corresponding keyword definition page. See #408 for more information. Since there are over 1000 keywords, and since I am not very confident that it will work correctly on every file, the plan is to apply the script to some keywords at a time. For example, starting with the 168 keywords in chapter 5. After it has been applied to all keyword files, we can consider applying it to the chapters and appendices also.


hakonhagland:
The script also needs to handle recursive `<text:p>` tags. This is an issue if a paragraph contains footnotes that are also marked up with `<text:p>`.

gdfldm:
Where a "keyword" appears multiple times in a KEYWORD.fodt file do we want add links to every occurrence or perhaps just the first?

hakonhagland:
> do we want add links to every occurrence or perhaps just the first?

@gdfldm We could also think of the links as a special markup for the keywords. Currently, the user recognizes a keyword by being an uppercase word, but there are upper case words that are not a keyword. However, if the keyword is also displayed in in e.g. italics and in a different color, the user will more easily recognize a keyword. Here is an example for how it will look for the `H2STORE` keyword:
 
image

Note that currently the keyword itself `H2STORE` is not marked up, but we could change that such that (it could link to itself) to avoid confusion (whether the `H2STORE` a keyword or not).

hakonhagland:
Currently, the script misses keywords that are split by span tags, for example if COSTORE occurs like this within the text `<text:span text:style-name="T6">C</text:span>O2STORE`

hakonhagland:
Rebased this onto #410

hakonhagland:
Rebased this and added a commit that will generate the mapping from keywords to URIs on the fly, see discussion in https://github.com/OPM/opm-reference-manual/pull/409#discussion_r1829463086 for more information

hakonhagland:
Updated script to handle aliases, see https://github.com/OPM/opm-reference-manual/pull/411#issuecomment-2465607885 for more information

blattms:
Seems like this needs a rebase, sorry.

hakonhagland:
> Hopefully, the number changed files is smaller after this. 

Actually this PR includes #417, so maybe we should merge that first?  @gdfldm can you have a look at this: https://github.com/OPM/opm-reference-manual/pull/417#issuecomment-2468170738

Then, I can rebase this one after that one has been merged.

blattms:
Ok the other PR is merged. Please rebase this and I hope that @lisajulia will take another look at the Python code.

hakonhagland:
> Ok the other PR is merged. Please rebase this 

@blattms Thanks! I rebased this now.

parts/meta/kw_uri_map.txt
```diff
@@ -201,8 +201,20 @@ ENDNUM __RefHeading___Toc123125_83452205
 ENDSCALE __RefHeading___Toc68146_2267116897
 ENDSKIP __RefHeading___Toc605472_3199477706
 ENKRVD __RefHeading___Toc69787_621662414
+ENKRVDX __RefHeading___Toc69787_621662414
```
blattms, line 204: Thanks. The changes to parts/meta/kw_uri_map.txt are intended?

hakonhagland: Yes. These are inserted by the alias functions. See https://github.com/OPM/opm-reference-manual/pull/410/files#diff-45a6fd9dbaf527c01ddb0d629af97c5b4ecf940c546aa9f3e834044d1985c6dbR135. They create aliases to keywords that are included in another keyword. See https://github.com/OPM/opm-reference-manual/pull/411#issuecomment-2465607885 for more information

hakonhagland:
> I suggest to name the files differently

@lisajulia Thanks for the review :) I have renamed the files in the last update.

PR #411, opened 2024-11-05
Added links for keyword files in chapter 5.3

hakonhagland:
Used the script in #410 to add links to all keyword files in Chapter 5.3. See #410 for more information.


parts/chapters/subsections/5.3/BRINE.fodt
```diff
@@ -4737,15 +4737,15 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
     <text:h text:style-name="Heading_20_4" text:outline-level="4"><text:bookmark-start text:name="__RefHeading___Toc715589_3964674244"/>Example<text:bookmark-end text:name="__RefHeading___Toc715589_3964674244"/></text:h>
     <text:p text:style-name="P64">The first example actives the standard Brine model and has no terminating “/”.</text:p>
     <text:p text:style-name="_40_Example">--</text:p>
-    <text:p text:style-name="_40_Example">-- <text:s text:c="6"/>ACTIVATE STANDARD BRINE MODEL IN THE RUN</text:p>
+    <text:p text:style-name="_40_Example">-- <text:s text:c="6"/>ACTIVATE STANDARD <text:a xlink:href="#__RefHeading___Toc162083_289573908">BRINE</text:a> MODEL IN THE RUN</text:p>
     <text:p text:style-name="_40_Example">--</text:p>
-    <text:p text:style-name="_40_Example">BRINE</text:p>
```
gdfldm, line 4748: @hakonhagland, it looks like a link has been created here possibly because the style-name was directly specified as "_40_Example" rather than indirectly as for example "P64".

parts/chapters/subsections/5.3/CO2SOL.fodt
```diff
@@ -4203,20 +4202,20 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
-    <text:p text:style-name="P52">The DISGASW keyword in the RUNSPEC section can be used to model dissolution of CO<text:span text:style-name="T20">2</text:span> in the Brine. <text:span text:style-name="T21">Note that hydrocarbon gas is not allowed to dissolve in the Brine (this is considered a reasonable assumption for most hydrocarbon gases).</text:span></text:p>
-    <text:p text:style-name="P52">The hydrocarbon <text:span text:style-name="T23">reservoir fluid</text:span> properties should be set up as normal apart from the water PVT properties, which will be defined by the simulators’ CO<text:span text:style-name="T20">2</text:span>-Brine PVT model, so the water PVT keywords (e.g. PVTW) are not required. The CO<text:span text:style-name="T20">2</text:span> PVT properties will also be defined by the CO<text:span text:style-name="T20">2</text:span>-Brine PVT model so the solvent PVT keywords (e.g. SDENSITY, PVTSOL) are also not required.</text:p>
```
gdfldm, line 4212: PVTSOL has not been converted to a link. This seems to be due to a typo in the file name PVSOL.fodt not PVTSOL.fodt.

parts/chapters/subsections/5.3/COMPS.fodt
```diff
@@ -4285,7 +4285,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P3161">1</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table2074.B2" office:value-type="string">
-       <text:p text:style-name="P5436">COMPS</text:p>
+       <text:p text:style-name="P5436"><text:a xlink:href="#__RefHeading___Toc27871_3671211675 Copy 1">COMPS</text:a></text:p>
```
gdfldm, line 4288: Here the parameter name COMPS has been interpreted as the keyword COMPS.

parts/chapters/subsections/5.3/INSPEC.fodt
```diff
@@ -4089,9 +4089,9 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
...ng___Toc45789_719036256">INIT</text:a> Index file can either be written out in formatted form as ASCII i.e. text files, if the <text:a xlink:href="#__RefHeading___Toc46651_1640804870">FMTOUT</text:a> keyword has been activated (*.FINSPEC), or binary format (*.<text:a xlink:href="#__RefHeading___Toc212474_2369005893">INSPEC</text:a>) if the <text:a xlink:href="#__RefHeading___Toc46651_1640804870">FMTOUT</text:a> keyword has not been activated. If the <text:a xlink:href="#__RefHeading___Toc45789_719036256">INIT</text:a> keyword in the <text:a xlink:href="#__RefHeading___Toc38674_784232322">GRID</text:a> section has been used to switch on the writing of the *. <text:a xlink:href="#__RefHeading___Toc45789_719036256">INIT</text:a> data file then a binary <text:a xlink:href="#__RefHeading___Toc45789_719036256">INIT</text:a> Index file is automatically written out as well, unless the <text:a xlink:href="#__RefHeading___Toc70892_4106839650">NOINSPEC</text:a> keyword in the <text:a xlink:href="#__RefHeading___Toc55591_1778172979">RUNSPEC</text:a> section has been used to switch off the writing of the <text:a xlink:href="#__RefHeading___Toc45789_719036256">INIT</text:a> Index file. </text:p>
```
gdfldm, line 4093: File name extensions can be mistaken for keywords, for example *.INSPEC.

parts/chapters/subsections/5.3/PATHS.fodt
```diff
@@ -4316,7 +4316,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
     <text:p text:style-name="P13694">/ <text:s text:c="6"/></text:p>
     <text:p text:style-name="P13694"/>
-    <text:p text:style-name="P11188">The above example defines “GRID” and “SCHD” aliases in the RUNSPEC section than can be used in the GRID and SCHEDULE sections of the input deck. The next example shows how to use the “GRID” alias with the INCLUDE keyword in the GRID section.</text:p>
+    <text:p text:style-name="P11188">The above example defines “<text:a xlink:href="#__RefHeading___Toc38674_784232322">GRID</text:a>” and “SCHD” aliases in the <text:a xlink:href="#__RefHeading___Toc55591_1778172979">RUNSPEC</text:a> section than can be used in the <text:a xlink:href="#__RefHeading___Toc38674_784232322">GRID</text:a> and <text:a xlink:href="#__RefHeading___Toc43945_784232322">SCHEDULE</text:a> sections of the input deck. The next example shows how to use the “<text:a xlink:href="#__RefHeading___Toc38674_784232322">GRID</text:a>” alias with the <text:a xlink:href="#__RefHeading___Toc605478_3199477706">INCLUDE</text:a> keyword in the <text:a xlink:href="#__RefHeading___Toc38674_784232322">GRID</text:a> section.</text:p>
```
gdfldm, line 4319: The path alias "GRID" has been mistaken for a keyword. Perhaps keyword-like strings within quotes could be ignored?

gdfldm:
I have created a PDF of the Manual based on this PR and it looks very nice. I think there are just a few special cases that should be addressed.

parts/chapters/subsections/5.3/CO2SOL.fodt
hakonhagland: Thanks for spotting that, I have added #415 to fix the filename.

parts/chapters/subsections/5.3/COMPS.fodt
hakonhagland: I do not immediately see how to fix this with the current algorithm. However, we could handle special cases like this manually by inserting a marker that the script can recognize. For example, inserting a span tag with a non-existing `text:style-name`. The script can then check for that style name.  Like this: `<text:span text:style-name="NOT_KEYWORD">COMPS</text:span>`, where the style "NOT_KEYWORD" indicates to script that it should not link this word.

parts/chapters/subsections/5.3/ACTDIMS.fodt
```diff
@@ -4328,17 +4328,17 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:covered-table-cell/>
      </table:table-row>
     </table:table>
-    <text:p text:style-name="P14560">Table <text:sequence text:ref-name="refTable18" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">5.3</text:sequence>: ACTDIMS Keyword Description</text:p>
+    <text:p text:style-name="P14560">Table <text:sequence text:ref-name="refTable18" text:name="Table" text:formula="ooow:Table+1" style:num-format="1">5.3</text:sequence>: <text:a xlink:href="#__RefHeading___Toc4408_421927891">ACTDIMS</text:a> Keyword Description</text:p>
```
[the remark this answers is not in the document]
hakonhagland: Might be easiest to do this using the manual marker method. See https://github.com/OPM/opm-reference-manual/pull/411#discussion_r1833325909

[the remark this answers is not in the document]
hakonhagland: There seems to be a lot more tables than figures. I think we might be able to handle the tables using the script. I will have a look at it.

hakonhagland:
@gdfldm Thanks for the detailed review! I have tried to address your comment in the latest commit.

gdfldm:
@hakonhagland Thank you for all the updates. I have recreated the manual and found some places in Section 5 where we might want to mark up the documents by hand to avoid generating links but otherwise it looks good:

- API: "API Tracking", "API gravity"
- CPR: "CPR linear solver", "CPR preconditioner"
- ENDSCALE: "SWL arrays" this is not referring to a keyword but the link may still be helpful?
- EOS: "Equation of State (EOS)", "EOS region"
- EQLOPTS: THPRES parameter is also a keyword
- GIMODEL: "GI option"
- NOINSPEC: "INIT index file"
- REGDIMS: "FIPNUM regions" this is not referring to a keyword but the link may still be helpful?
- SATOPTS: A link has been generated in the title of the second Table in the description of this keyword.
- SAVE: "SAVE file"
- VE: "VE model"
- VISAGE: "VISAGE option" could simply change to "Visage option"?

hakonhagland:
>  found some places in Section 5 where we might want to mark up the documents by hand

@gdfldm Thanks! I checked the `SATOPTS` page now, and saw that `KRNUMX` and other similar keywords was not recognized. I think that is because they are not separate keywords in the manual. They are all described under the  `KRNUM` keyword. Maybe the script should set up aliases for these?

gdfldm:
> Maybe the script should set up aliases for these?

Yes, I think so.

hakonhagland:
@gdfldm See the last commit for updates to address the comments above.

parts/chapters/subsections/5.3/SATOPTS.fodt
gdfldm, line 1: For keywords like KRNUMX- the "-" seems to be placed outside the tag, for example:
```<text:a xlink:href="#__RefHeading___Toc273792_2369005893">KRNUMX</text:a>-, ```

hakonhagland:
> For keywords like KRNUMX- the "-" seems to be placed outside the tag

@gdfldm Good observation! I have fixed the script to handle this case. Check the latest commit for the result.

hakonhagland:
Rebased

PR #417, opened 2024-11-07
Insert special markers around words to avoid linking

hakonhagland:
Currently only `COMPS` in section 5.3 is marked. See: https://github.com/OPM/opm-reference-manual/pull/411#discussion_r1832985811. I plan to update this PR when more cases are spotted.


blattms:
I think @gdfldm should comment on this. It might be that he only had COMPS in the table in mind, where the meaning was different for sure.

I am not 100% sure whether I understand and agree with all the changes. Especially, CPR is keyword and the links would be nice at most places. For API it is a bit different, but IMHO, if it says "API tracking" it would be nice to have a link. 

Anyway, a few false links might not even hurt too much... Just my 2 cents.

gdfldm:
I agree generally it would be good to have a link anyway. For example, when referring to the PERMX array a link to the PERMX keyword could be helpful. 
In the specific cases here of API and CPR these all appear on the keyword page so I don't think the link adds anything, for example:
image
Where "API tracking" appears elsewhere in the manual I think it is worth having a link.

blattms:
@hakonhagland What do you think? Will you add changes?

Maybe we should not do to much special-casing in the scripts to keep them simple. may a link to the current subsection or one that points to a page that is a bit unrelated does not hurt that much to warrant extra work. Our readers are smart people and probably won't click on those.

hakonhagland:
> Will you add changes?

@blattms I think we can proceed with the changes in this PR. These changes are in the `.fodt` files themselves, so it does not complicate the script to have them or not to have them. I think as you said: It does not matter much if we have these changes (avoid linking in the special cases listed here) or not. But as @gdfldm has read through Chapter 5 more carefully in his review, I trust that these changes (not linking for the special cases listed here) will improve the reader experience.

PR #431, opened 2024-12-06
Added links for keywords in chapter 10.3

hakonhagland:


parts/chapters/subsections/10.3/APIVD.fodt
gdfldm, line 1: I think it would be good if some of the links to API could be systematically removed, for example the units degrees API "oAPI" and where the text is subscripted (gamma_API)?

parts/chapters/subsections/10.3/AQANCONL.fodt
gdfldm, line 1: Could we add a check for `&quot;` ?

parts/chapters/subsections/10.3/AQUFET.fodt
gdfldm, line 1: I think it would be best to remove the links to the DATUM keyword in the aquifer keyword descriptions, since the DATUM keyword refers to the model datum whereas the DATUM parameter is the aquifer datum?

parts/chapters/subsections/10.3/AQUFLUX.fodt
gdfldm, line 1: Perhaps remove links to TEMP keyword as this refers to a different temperature than TEMP parameter?

parts/chapters/subsections/10.3/BC.fodt
gdfldm, line 1: Remove link from TEMP parameter?

parts/chapters/subsections/10.3/EQUIL.fodt
gdfldm, line 1: Perhaps remove link to DATUM keyword?

parts/chapters/subsections/10.3/AQANCONL.fodt
hakonhagland: Yes good idea. Later in that section there is also a lot of `LGR's` where the apostrophe is encoded with
`LGR</text:a>&apos;`. Maybe we should avoid linking these keywords also?

gdfldm:
I have created the reference manual PDF and think the updates to Chapter 10 look very good. In my opinion some of the parameter names that are the same as keyword names would be better without links, but don't think this needs to be done before merging this PR.

parts/chapters/subsections/10.3/AQANCONL.fodt
hakonhagland: There are some technical issues here. xml.sax parser is splitting these quotes and apostrophes into separate invocations of the callback. I will have to think a little bit more on this.

hakonhagland: Found a workaround, see #433.

hakonhagland:
@gdfldm Thanks for reviewing. I have added some commits to address your comments.

blattms:
As @gdfldm has done the reviewing, he should merge this whenever seems this to be good. I won't invest time here.

PR #432, opened 2024-12-10
Update DRSDTCON keyword with regimes and option for GAS/WATER

gdfldm:
Extended DRSDTCON with regimes, and option for GAS/WATER
See https://github.com/OPM/opm-simulators/pull/5491


blattms:
We might have a race-condition with #435. If that is merged first this one might have conflicts. Would that be a problem for you @gdfldm?

gdfldm:
> Would that be a problem

I think it would be ok.

blattms:
Thanks.
Unfortunately, this now needs a rebase.

gdfldm:
> this now needs a rebase.

I think all of the conflicts have now been resolved

blattms:
Thanks. Let's see what @totto82 has to say.

totto82:
Thanks, it mostly looks good. 

You write: "CHI typically varies from 0.01 -  0.1 depending on reservoir properties. " 

The way the model is formulated, CHI and PSI are independent on reservoir properties by definition. (Omega is probably not, but also less important). Fine-scale simulations indicates that 0.04+-0.01 are reasonable numbers for CHI. The idea is that the user does not need to do fine-scale simulations themselves but can use the numbers reported in the paper. So 0.04 is a good default and the mean value given in the paper. 

Where did you get the number 0.01 to 0.1 from? 

I also notices a spelling mistake "dissoved"

gdfldm:
@totto82. Thanks for the review. 
I have removed the typical range, and expanded the discussion of the CHI value.

totto82:
> This came from the previous definition of DRSDTCON in an earlier version of the manual

Ok. Then I probably added it myself. In the updated model this is a too large range. Thanks for the update. From my point of view this can me merged.

totto82 (approved):
Thanks for the update.

PR #433, opened 2024-12-10
Extend keyword linking script to handle quoted keywords

hakonhagland:
Extends the script to handle keywords in quotes represented by an `&quot;` entity or keywords followed by apostrophes represented by an `&apos;` entity. We should not link these keywords.


lisajulia:
I've looked at this PR and tested "fodt-link-keywords" locally and it works for me - so I'll merge this now.

PR #434, opened 2024-12-10
Added links for keywords in chapter 11.3

hakonhagland:


parts/chapters/subsections/11.3/EXCEL.fodt
```diff
@@ -4100,7 +4100,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
      </table:table-row>
     </table:table>
     <text:h text:style-name="P18345" text:outline-level="4" text:is-list-header="true"><text:bookmark-start text:name="__RefHeading___Toc10367_370116838819131"/>Description<text:bookmark-end text:name="__RefHeading___Toc10367_370116838819131"/></text:h>
-    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the RSM file data in a format that can easily be loaded into Microsoft&apos;s EXCEL spreadsheet program or LibreOffice’s CALC spreadsheet program. The RSM file output is activated by the RUNSUM keyword in the SUMMARY section.</text:p>
+    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the RSM file data in a format that can easily be loaded into Microsoft&apos;s <text:a xlink:href="#__RefHeading___Toc210148_2884651453">EXCEL</text:a> spreadsheet program or LibreOffice’s CALC spreadsheet program. The RSM file output is activated by the <text:a xlink:href="#__RefHeading___Toc210156_2884651453">RUNSUM</text:a> keyword in the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> section.</text:p>
```
blattms, line 4103: EXCEL might be a keyword, but here it a brand of a software provider. We should probably change the case.

parts/chapters/subsections/11.3/EXCEL.fodt
```diff
@@ -4119,7 +4119,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
     <text:p text:style-name="P14099">--</text:p>
     <text:p text:style-name="P13192">EXCEL <text:s text:c="83"/><text:s text:c="4"/></text:p>
     <text:p text:style-name="P15481"/>
-    <text:p text:style-name="P15492">The above example activates the SUMMARY file EXCEL option for directly loading the RSM file into either Microsoft&apos;s EXCEL or LibreOffice’s CALC spreadsheet programs.</text:p>
+    <text:p text:style-name="P15492">The above example activates the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> file <text:a xlink:href="#__RefHeading___Toc210148_2884651453">EXCEL</text:a> option for directly loading the RSM file into either Microsoft&apos;s <text:a xlink:href="#__RefHeading___Toc210148_2884651453">EXCEL</text:a> or LibreOffice’s CALC spreadsheet programs.</text:p>
```
blattms, line 4122: same here.

parts/chapters/subsections/11.3/OFM.fodt
```diff
@@ -4076,9 +4076,9 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
      </table:table-row>
     </table:table>
     <text:h text:style-name="P18345" text:outline-level="4" text:is-list-header="true"><text:bookmark-start text:name="__RefHeading___Toc10367_3701168388191311112"/>Description<text:bookmark-end text:name="__RefHeading___Toc10367_3701168388191311112"/></text:h>
-    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the SUMMARY file data in the Oil Field Manager (“OFM”) file format to enable the simulated data to be directly loaded into OFM. </text:p>
+    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> file data in the Oil Field Manager (“OFM”) file format to enable the simulated data to be directly loaded into <text:a xlink:href="#__RefHeading___Toc234776_2928331029">OFM</text:a>. </text:p>
```
blattms, line 4079: OFM is not used as a keyword here. Maybe rewrite as "Oil Field Manager"?

parts/chapters/subsections/11.3/RUNSUM.fodt
```diff
@@ -4095,9 +4095,9 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
-    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the SUMMARY file data in a columnar format to the PRT file. Normally the SEPARATE keyword in the SUMMARY section is invoked in the same run to direct the data stream to a separate RSM file for easy loading into other programs, for example, Microsoft&apos;s EXCEL or LibreOffice’s CALC spreadsheet programs.</text:p>
+    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> file data in a columnar format to the PRT file. Normally the <text:a xlink:href="#__RefHeading___Toc210158_2884651453">SEPARATE</text:a> keyword in the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> section is invoked in the same run to direct the data stream to a separate RSM file for easy loading into other programs, for example, Microsoft&apos;s <text:a xlink:href="#__RefHeading___Toc210148_2884651453">EXCEL</text:a> or LibreOffice’s CALC spreadsheet programs.</text:p>
```
blattms, line 4098: EXCEL (MS version again

parts/chapters/subsections/11.3/SEPARATE.fodt
```diff
@@ -4101,9 +4101,9 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+    <text:p text:style-name="_40_TextBody">This keyword activates the writing out of the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> file date in a columnar format to the RSM file, if the <text:a xlink:href="#__RefHeading___Toc210156_2884651453">RUNSUM</text:a> keyword has also been activated in the <text:a xlink:href="#__RefHeading___Toc43949_784232322">SUMMARY</text:a> section. Both the <text:a xlink:href="#__RefHeading___Toc210158_2884651453">SEPARATE</text:a> and the <text:a xlink:href="#__RefHeading___Toc210156_2884651453">RUNSUM</text:a> keywords need to be invoked. If the <text:a xlink:href="#__RefHeading___Toc210158_2884651453">SEPARATE</text:a> option is not activated then the RSM output is directed to the end of the PRT file. Normally the both the <text:a xlink:href="#__RefHeading___Toc210158_2884651453">SEPARATE</text:a> and <text:a xlink:href="#__RefHeading___Toc210156_2884651453">RUNSUM</text:a> keywords are invoked in the same run to enable easy loading of the data into Microsoft&apos;s <text:a xlink:href="#__RefHeading___Toc210148_2884651453">EXCEL</text:a> or LibreOffice’s CALC spreadsheet programs.</text:p>
```
blattms, line 4104: EXCEL

blattms (commented):
Looks mostly perfect.

Just a few false positives to fix.

hakonhagland:
@blattms Thanks for the review :) I have added some commits to address the comments.

blattms:
So we are using `<text:span text:style-name="NOT_KEYWORD">..</text: text:style-name="NOT_KEYWORD"> to prevent the script from translating it. Is that really persistent? If I ever open that file with libreoffice and save again then seems to be gone.
So we maybe need another way?

hakonhagland:
> Is that really persistent?

@blattms Good observation. Unfortunately, it seems it is not persistent. I assumed that libreoffice did not clean it up if it did not recognize. Hmm.. Maybe we need to predefine it for libreoffice to accept it? I am investigating `Styles > Manage Styles` if that can work.

gdfldm:
It seems to be possible to trick LO by adding something like...
```
  <style:style style:name="Not_20_Keyword" style:display-name="Not Keyword" style:family="text">
   <style:text-properties fo:language="en" fo:country="US"/>
  </style:style>
```

blattms:
I see several hacks (at least for this one:
- Use Excel
- Put it into double quotes
- Add ugly whitespace using `<text:s/>` around it. Your eyes won't see it. I came across extra whitespace in many reviews.

Maybe using an extra style that is a copy of the other might work, too.

As I might have said before: False positives are really not that bad. Engineers should be smart enough to recognize these and not click on links...

hakonhagland:
> It seems to be possible to trick LO

 @gdfldm  Yes that seems to work :) I tried inserting `<style:style style:name="NOT_KEYWORD" style:display-name="Not a keyword" style:family="text"/>` into the `<office:styles>` section. This work almost, except that when I saved the document libreoffice changed the style name to `Not_20_a_20_keyword`, so it became like this:

`<style:style style:name="Not_20_a_20_keyword" style:display-name="Not a keyword" style:family="text"/>`

and `<text:span text:style-name="Not_20_a_20_keyword">EXCEL</text:span>` this..

gdfldm:
It looks like the style name needs to be consistent with the display name, with the spaces converted to _20_

hakonhagland:
Tried this variation:
` <style:style style:name="NOT_KEYWORD" style:display-name="Not_keyword" style:family="text"/>`
then libreoffice converted the underline to `5f`, like this:
`<style:style style:name="Not_5f_keyword" style:display-name="Not_keyword" style:family="text"/>`

gdfldm:
Another option might be to only add a link to an occurrence of a KEYWORD where it is followed by "keyword", "section" or "array" in the same sentence (for example "WCONPROD keyword", "SCHED section", or "PERMX array") ?

gdfldm:
> Tried this variation:

Is that because you have an underscore in the display name?

hakonhagland:
> Is that because you have an underscore in the display name?

@gdfldm Yes, right. If I omit underscores and spaces and use the same style name as display name, I can even omit the display name from the style definition like this:

` <style:style style:name="NotKeyword" style:family="text"/>`

and it seems to work (libreoffice accepts it and does not change it)

hakonhagland:
I can write a script that will change the existing `NOT_KEYWORD` styles (in chapters 4-10) to `NotKeyword` and insert the style definition 

`<style:style style:name="NotKeyword" style:family="text"/>`

in the `<office:styles>` section. This should fix the problem. What do you think?

hakonhagland:
> So the plan is to merge this as it is now?

@blattms Yes good idea. Then the script will automatically fix chapter 11 after that.

gdfldm:
I have found that the style definition can be deleted/renamed if it appears after the start of the unnamed styles (P1, etc).

hakonhagland:
> style definition can be deleted/renamed if it appears after the start of the unnamed styles

@gdfldm Interesting observation. Does these unnamed styles correspond to xml section `<office:automatic-styles>` ? So we should not put the new `NotKeyword` style in that section, but rather in the `<office:styles>` section then?

gdfldm:
> So we should not put the new `NotKeyword` style in that section, but rather in the `<office:styles>` section then?

Yes, from my experiments it looks like it needs to be in `<office:styles>`

PR #436, opened 2024-12-12
Script to change span tag NOT_KEYWORD attribute to NotKeyword

hakonhagland:
Script that will change the span tag attribute `NOT_KEYWORD` to `NotKeyword`, and also add a corresponding style in the `office:styles` section. See https://github.com/OPM/opm-reference-manual/pull/434#issuecomment-2539457203 for more information


lisajulia:
@hakonhagland: When having merged #437, do we still need this script here? (In #437, the description says it only covers chapters 4-10... so then we do, right?)

lisajulia:
> @hakonhagland: When having merged #437, do we still need this script here? (In #437, the description says it only covers chapters 4-10... so then we do, right?)

So we need this PR, i.e. I will merge this now.

hakonhagland:
> I will merge this now.

@lisajulia Thank you for the review. Sorry I missed your comments above. Yes this script might be useful have for later.

PR #437, opened 2024-12-12
Changed span tag attribute NOT_KEYWORD to NotKeyword for some keyword files

hakonhagland:
Used the script in #436 to changed the span tag attribute `NOT_KEYWORD` to `NotKeyword` and also add a corresponding style in the `office:styles` section. This only includes keyword files in chapters 4-10. See https://github.com/OPM/opm-reference-manual/pull/434#issuecomment-2539457203 for more information.


hakonhagland:
We might want to install the style `NotKeyword`:

`<style:style style:name="NotKeyword" style:family="text"/>`

in all .fodt files even if they do not use the style. That would make it easier not to forget to add the style if we later add a span tag with `NotKeyword` to a new file.

blattms (approved):

blattms:
Thanks a lot.

PR #435, opened 2024-12-12
Added links for keywords in chapter 12.3

hakonhagland:

PR #439, opened 2024-12-17
Update link keywords script: Do not link keywords followed by a dash

hakonhagland:
Do not link keywords like "LGR" in the string "LGR-OPM1". See https://github.com/OPM/opm-reference-manual/pull/435#discussion_r1888930577 for more information


lisajulia:
tested locally, it works

PR #441, opened 2024-12-17
Update link keywords script: Do not add links for figure captions

hakonhagland:
Builds on #440 which should be merged first.

To avoid adding links from Figure captions, we check if we are inside a `draw:frame` element. See https://github.com/OPM/opm-reference-manual/pull/435#discussion_r1889142988 for more information


lisajulia:
Reviewed and tested, it works as intended :+1:

PR #442, opened 2024-12-19
Script to check for erroneous links

hakonhagland:
Builds on #441 which should be merged first.

This script checks for erroneously inserted links to keywords in the `office:binary-data` sections in the keyword files. These links might have been inserted by the keyword linking script, see https://github.com/OPM/opm-reference-manual/pull/435#issuecomment-2553264255 for more information.

I ran this script on chapters 4-11 and no erroneous links where detected there (according to this script). I will check chapter 12 next.

PR #444, opened 2024-12-20
Update keyword linking script to handle chapters

hakonhagland:
Builds on #442 which should be merged first.

Updated the keyword linking script to handle chapter `.fodt` files like `1.fodt`.


lisajulia:
Reviewed and tested: :+1:

PR #446, opened 2024-12-20
Added links for keywords in chapter 2

hakonhagland:


parts/chapters/2.fodt
```diff
@@ -88314,8 +88313,8 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
      <table:table-header-rows>
       <table:table-row table:style-name="Table2272.1">
        <table:table-cell table:style-name="Table2272.A1" office:value-type="string">
-        <text:p text:style-name="P603">Warning: VFPPROD table 39 has 5 non-monotonic points of BHP(THP)</text:p>
-        <text:p text:style-name="P603">In /media/OPM/Fields/Norne/Model/INCLUDE/VFP/B3H.Ecl line 68</text:p>
+        <text:p text:style-name="P603">Warning: <text:a xlink:href="#__RefHeading___Toc121919_2556401936">VFPPROD</text:a> table 39 has 5 non-monotonic points of BHP(THP)</text:p>
+        <text:p text:style-name="P603">In /media/OPM/Fields/Norne/Model/<text:a xlink:href="#__RefHeading___Toc605478_3199477706">INCLUDE</text:a>/VFP/B3H.Ecl line 68</text:p>
```
blattms, line 88317: Oops, there should be no link here. The script needs to detect paths.

parts/chapters/2.fodt
```diff
@@ -91748,13 +91747,13 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:table-row table:style-name="Table2273.1">
        <table:table-cell table:style-name="Table2273.A1" office:value-type="string">
         <text:p text:style-name="P603"/>
-        <text:p text:style-name="P603">Warning: LIQ preferred phase not supported for well OP-01, using OIL instead</text:p>
+        <text:p text:style-name="P603">Warning: LIQ preferred phase not supported for well OP-01, using <text:a xlink:href="#__RefHeading___Toc97439_1778172979">OIL</text:a> instead</text:p>
```
blattms, line 91750: I think we need to guard this one. It should not be a link.

blattms (commented):
I think the script needs to detect paths. It marked INCLUDE in a path.

parts/chapters/2.fodt
hakonhagland: Good catch. Can we check for style P603 here? It seem like both cases uses that style to produce a mono space font on a gray background

hakonhagland:
> I think we might need to guard this one. It should not be a link.

@blattms I used the updated script in #452 to handle those, see latest commit.

parts/chapters/2.fodt
blattms: Like Matthew I would assume that these numbers might change.

Proably, having an extra link is not such a problem. Hence we should not invest too much time.

PR #449, opened 2024-12-23
Update CNAMES keyword

gdfldm:
Three components H20, CO2 and NACL are now required if used with the CO2STORE model


parts/chapters/subsections/8.3/CNAMES.fodt
```diff
@@ -4264,8 +4267,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
-        <text:p text:style-name="P10209">Secondly, although OPM Flow parses the keyword, the simulator currently ignores the data for this keyword.</text:p>
+        <text:p text:style-name="P10195">This keyword is only supported by OPM Flow when the two phase gas-water CO<text:span text:style-name="T999">2</text:span> storage model has been activated using the <text:a xlink:href="#__RefHeading___Toc387968_1616145207">CO2STORE</text:a> keyword and either the <text:a xlink:href="#__RefHeading___Toc38607_2267116897 Copy 1">GASWAT</text:a> or the <text:a xlink:href="#__RefHeading___Toc38607_2267116897">GAS</text:a> and <text:a xlink:href="#__RefHeading___Toc38611_2267116897">WATER</text:a> keywords in the <text:a xlink:href="#__RefHeading___Toc55591_1778172979">RUNSPEC</text:a> section. The component names "H2O", "CO2" and "NACL" (water, CO<text:span text:style-name="T999">2</text:span> and salt respectively) must be included when CNAMES is used with the <text:a xlink:href="#__RefHeading___Toc387968_1616145207">CO2STORE</text:a> keyword.</text:p>
```
blattms, line 4270: I might be misunderstanding this.

To me it sounds like I have provide all three of those all the time, but it seems like I need to select only NCOMPS components from this set (H2O, CO2, NACL). At least there are cases in opm-tests/co2store that only use 2 (e.g. CO2STORE_GASWAT).

The new example below runs, but I would hope that the old version does, too.

gdfldm: As I understand it you need to specify H2O, CO2, NACL in order for the DENAQA and VISCAQA know which order the Erzokhi parameters are given in the tables.

gdfldm: I will check and update the text.

parts/chapters/subsections/8.3/CNAMES.fodt
```diff
@@ -4264,8 +4267,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+        <text:p text:style-name="P10195">This keyword is only supported by OPM Flow when the two phase gas-water CO<text:span text:style-name="T999">2</text:span> storage model has been activated using the <text:a xlink:href="#__RefHeading___Toc387968_1616145207">CO2STORE</text:a> keyword and either the <text:a xlink:href="#__RefHeading___Toc38607_2267116897 Copy 1">GASWAT</text:a> or the <text:a xlink:href="#__RefHeading___Toc38607_2267116897">GAS</text:a> and <text:a xlink:href="#__RefHeading___Toc38611_2267116897">WATER</text:a> keywords in the <text:a xlink:href="#__RefHeading___Toc55591_1778172979">RUNSPEC</text:a> section. Only the component names "H2O", "CO2" and "NACL" (water, CO<text:span text:style-name="T999">2</text:span> and salt respectively) are recognized when the <text:a xlink:href="#__RefHeading___Toc27871_3671211675 Copy 1 Copy 1 Copy 1">CNAMES</text:a> keyword is used with the <text:a xlink:href="#__RefHeading___Toc387968_1616145207">CO2STORE</text:a> keyword; any other component names are ignored.</text:p>
```
blattms, line 4270: Sorry to bother you again. What is up with the ` Copy 1 Copy 1 Copy1` in the href? Is it supposed to be like this?

blattms (changes requested):

parts/chapters/subsections/8.3/CNAMES.fodt
gdfldm: No problem. All feedback is appreciated :).
I think these are created when a bookmark is copied and pasted in LibreOffice.
I will try and replace it with something better.

gdfldm: Changed to:
REF_HEADING_KEYWORD_CNAMES_8_3

gdfldm: > have more those

I have found about 18 keywords and 15 headings that are similar.

blattms: And no it turns out that my comment was bad and might have made things worse:

Indeed those bookmarks are from the original document.They first appeared in 3b1160e46c0d. Hence maybe it is safer to leave them as they are and revert/remove the last commit?

Sorry again...

blattms: > But it does seem to be taking longer to do Tools>Update>Update All in LibreOffice?

Probably that changes it again, because it does not adhere to what libreoffice expects?

blattms (approved):

PR #452, opened 2025-01-03
Update keyword linking script to handle P603 paragraphs

hakonhagland:
These paragraphs contain verbatim monospaced text and should not be linked. See https://github.com/OPM/opm-reference-manual/pull/446#discussion_r1901703791 for more information

### DIFF
--- a/scripts/python/src/fodt/keyword_linker.py
+++ b/scripts/python/src/fodt/keyword_linker.py
@@ -7,6 +7,7 @@ import xml.sax.handler
 import xml.sax.xmlreader
 import xml.sax.saxutils
 
+from dataclasses import dataclass
 from pathlib import Path
 
 import click
@@ -19,6 +20,26 @@ from fodt.xml_helpers import XMLHelper
 class FileType(enum.Enum):
     CHAPTER = 1
     SUBSECTION = 2
+
+@dataclass
+class MonoParagraphStyle:
+    style_name: str = ""
+    loext_graphic_properties: bool = False
+    style_paragraph_properties: bool = False
+    style_text_properties: bool = False
+    libre_mono_font: bool = False
+    libre_mono_font_size: bool = False
+
+    def in_style_style_element(self) -> bool:
+        return self.style_name != ""
+
+    def valid(self) -> bool:
+        return (self.loext_graphic_properties
+                and self.style_paragraph_properties
+                and self.style_text_properties
+                and self.libre_mono_font
+                and self.libre_mono_font_size)
+
 
 class FileHandler(xml.sax.handler.ContentHandler):
     def __init__(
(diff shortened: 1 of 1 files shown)


gdfldm:
What if LibreOffice changes all of the style numbers?

hakonhagland:
> What if LibreOffice changes all of the style numbers?

@gdfldm Good point. The style P603 is defined like this:

 ```
 <style:style style:name="P603" style:family="paragraph" style:parent-style-name="Text_20_body">
   <loext:graphic-properties draw:fill="none" draw:fill-gradient-name="gradient" draw:fill-hatch-name="hatch"/>
   <style:paragraph-properties fo:margin-left="0cm" fo:margin-right="0.318cm" fo:margin-top="0cm" fo:margin-bottom="0.051cm" style:contextual-spacing="false" fo:text-align="start" style:justify-single-word="false" fo:orphans="0" fo:widows="0" fo:hyphenation-ladder-count="no-limit" fo:text-indent="0cm" style:auto-text-indent="false" fo:background-color="transparent" style:shadow="none" style:writing-mode="lr-tb"/>
   <style:text-properties style:font-name="Liberation Mono" fo:font-size="8pt" style:font-size-asian="8pt" style:font-size-complex="8pt" fo:hyphenate="false" fo:hyphenation-remain-char-count="2" fo:hyphenation-push-char-count="2" loext:hyphenation-no-caps="false" loext:hyphenation-no-last-word="false" loext:hyphenation-word-char-count="5" loext:hyphenation-zone="no-limit"/>
  </style:style>
```
The script could search through all automatic-styles and determine if the is one or more styles that has for example:

- style:family="paragraph" 
- style:parent-style-name="Text_20_body"
- loext:graphic-properties sub element
- style:text-properties sub element with
  - style:font-name="Liberation Mono"
  - fo:font-size="8pt"

and record the value of style:name (here P603) for such styles, or alternatively we could create a new custom style as we did with the `NotKeyword` style, and but that in the `office:styles` section (and not in `office:automatic-styles` as it is now)

Any other ideas for how to solve this?

**Comment by hakonhagland** [comment]
> and record the value of style:name (here P603) for such styles

I chose to try out this approach first, see latest commit.

**Comment by hakonhagland** [comment]
> Maybe in addition detect keywords in a path (between slashes or between slash and dot)?

@blattms Yes this is a good idea, but I suggest we delay the feature unless we have an explicit case that shows up where it is [needed](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

File: scripts/python/src/fodt/keyword_linker.py [hunk 1]
```diff
@@ -20,6 +21,26 @@ class FileType(enum.Enum):
...
+
+    def in_style_style_element(self) -> bool:
+        return self.style_name != ""
+
+    def valid(self) -> bool:
+        return (self.loext_graphic_properties
+                and self.style_paragraph_properties
+                and self.style_text_properties
+                and self.libre_mono_font
+                and self.libre_mono_font_size)
```
lisajulia, line 41: @hakonhagland: That then means the font size needs to be set to exactly 8pt such that the MonoParagraphStyle is valid, right? How about other font sizes?

lisajulia:
I generally approve this PR :+1:  :), yet I have one remaining question: https://github.com/OPM/opm-reference-manual/pull/452#discussion_r1904921965, as soon as that is answered, I'm happy to merge this.

scripts/python/src/fodt/keyword_linker.py
hakonhagland: Right, the font size might be changed in the future. I suggest we allow any font size to be included in the MonoParagraphStyle

hakonhagland:
@lisajulia Thanks for the review. I have a addressed the comment in the latest commit

lisajulia:
> @lisajulia Thanks for the review. I have a addressed the comment in the latest commit

Thanks :) yet I'm wondering if the check for the font size could be removed overall, since we check if style:font-name is Liberation Mono? Sorry for the nitpicking...

hakonhagland:
@lisajulia Removed the font size check.

PR #455, opened 2025-01-03
Added links for keywords in chapter 3

hakonhagland:


blattms (approved):
Thanks, I'll merge. This chapter is more a description of the style of the keyword description with examples. But links won't hurt even if they won't be clicked.

PR #456, opened 2025-01-03
Added links for keywords in chapter 4

hakonhagland:


blattms:
Hey, how many chapters do we have left without links?

hakonhagland:
> how many chapters do we have left without links?

@blattms Chapters 5-12 and appendices A-F

gdfldm:
> appendices A-F

I think we can skip Appendix A

blattms:
> > appendices A-F
> 
> I think we can skip Appendix A

And C (and maybe even more).

hakonhagland:
> A and C (and maybe even more)

@blattms I agree that we can skip A, why do we want to skip any of the others? I suggest we check how the script performs on them first

blattms:
C is a about the python program opmrun.

hakonhagland:
> C is a about the python program opmrun.

Yes but it still references keywords. I tested the script on Appendix C now, see #467.

PR #458, opened 2025-01-06
Add SALTMF keyword

gdfldm:


blattms:
@hakonhagland Now that we have all the links, should we proceed when adding a keyword.

Does the add-keyword script need to modify the parts/meta/kw_uri_map.txt file? Do we need to run a script that adds possible new links?

blattms (approved):
Just waiting for @hakonhagland's reply

hakonhagland:
> Does the add-keyword script need to modify the parts/meta/kw_uri_map.txt file? Do we need to run a script that adds possible new links?

@blattms Yes :) We should run the script `fodt-gen-kw-uri-map` to update the file `parts/meta/kw_uri_map.txt` and run the script `fodt-link-keywords` to create potential new links. It would be nice if we could automate this process somehow..

hakonhagland:
> It would be nice if we could automate this process somehow..

Maybe we can set up a GitHub action workflow that runs the two scripts when a PR is submitted. If the scripts detects a change to either `parts/meta/kw_uri_map.txt` or to any `.fodt` file, the workflow will fail and it will not be possible to merge the PR. What do you think?

blattms:
Automatically creating a PR just for changes in parts/meta/kw_uri_map.txt might make sense.

PR #464, opened 2025-01-07
Added links for keywords in chapters 8-12

hakonhagland:


parts/chapters/10.fodt
```diff
@@ -20742,7 +20742,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
                   <mspace width="2em"/>
                  </mrow>
                  <msub>
-                  <mi>&#x03C1;</mi>
+                  <mi>ρ</mi>
```
blattms, line 20745: This change is interesting. Why wasn't this done during the split?
But it works.

blattms (approved):
Quickly checked. (I have lost steam)

parts/chapters/10.fodt
hakonhagland: I think this is done automatically by the python `xml.sax` parser, when we write it back to file the `xml.sax.saxutils.escape()` does not convert back to the unicode hex representation. This is fine since our xml files are using UTF-8 encoding, the first line in the `.fodt` files is like this:

```
<?xml version="1.0" encoding="UTF-8"?>
```

so it should work fine. The reason the original splitting did not do this conversion might be due to higher level parsing of the sections in Python (and the xml.sax parser was never called on this text)

gdfldm: This might have been me inserting a rho in the text editor?

PR #467, opened 2025-01-07
Added links for keywords in appendix C

hakonhagland:


blattms:
Indeed there were a few  keywords here :sweat_smile: and no issues :smile:

PR #469, opened 2025-01-08
Added links for keywords in appendix E

hakonhagland:


parts/appendices/E.fodt
```diff
@@ -17694,7 +17693,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="_40_Table_20_Contents"/>
        <text:p text:style-name="_40_Table_20_Contents">Where:</text:p>
        <text:p text:style-name="P911">DECK is the name of the data deck you want to get the time steps from,</text:p>
-       <text:p text:style-name="P911">TIME tells the application to return the timing for the run, and</text:p>
+       <text:p text:style-name="P911"><text:a xlink:href="#__RefHeading___Toc1252966_4250154414">TIME</text:a> tells the application to return the timing for the run, and</text:p>
```
blattms, line 17696: Not used as a keyword here. Maybe remove link?

parts/appendices/E.fodt
```diff
@@ -19875,7 +19874,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P822"/>
        <text:p text:style-name="P822">Where:</text:p>
        <text:p text:style-name="P825">DECK is the name of the data deck you want to get the time steps from,</text:p>
-       <text:p text:style-name="P825">TIME tells the application to return the timing for the run, and</text:p>
+       <text:p text:style-name="P825"><text:a xlink:href="#__RefHeading___Toc1252966_4250154414">TIME</text:a> tells the application to return the timing for the run, and</text:p>
```
blattms, line 19877: TIME is not used as a keyword here?

Out of curiosity: Why did the script not replace it in line 10874? There it keeps
```
<text:p text:style-name="P610">path_to_libecl_applications/ecl_summary DECK TIME &gt; filename</text:p>
```

parts/appendices/E.fodt
```diff
@@ -21968,7 +21967,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P778"/>
        <text:p text:style-name="P778">Where:</text:p>
        <text:p text:style-name="P788">DECK is the name of the data deck you want to get the time steps from,</text:p>
-       <text:p text:style-name="P788">TIME tells the application to return the timing for the run, and</text:p>
+       <text:p text:style-name="P788"><text:a xlink:href="#__RefHeading___Toc1252966_4250154414">TIME</text:a> tells the application to return the timing for the run, and</text:p>
```
blattms, line 21970: Lot's of these, it seems.

parts/appendices/E.fodt
```diff
@@ -24023,7 +24022,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P609">path_to_libecl_applications/ecl_summary DECK TIME &gt; filename</text:p>
        <text:p text:style-name="P761">Where:</text:p>
        <text:p text:style-name="P770">DECK is the name of the data deck you want to get the time steps from,</text:p>
-       <text:p text:style-name="P770">TIME tells the application to return the timing for the run, and</text:p>
+       <text:p text:style-name="P770"><text:a xlink:href="#__RefHeading___Toc1252966_4250154414">TIME</text:a> tells the application to return the timing for the run, and</text:p>
```
blattms, line 24025: Not used as keyword.
There seem to be more occurences below. I'll stop commenting about those from here.

blattms (commented):
Not 100% sure what to do here.
All occurence of TIME are not as a keyword. 

I am wondering  whether that links actually hurts. On the other hand one might wonder whether we need links in this appendix.

If it is not too much effort I would just make that links for TIME are skipped. Otherwise we live with them. Your call.

parts/appendices/E.fodt
hakonhagland: > Out of curiosity: Why did the script not replace it in line 10874? 

It is because it has a style P610 which is a sub style of the Example style, see

https://github.com/OPM/opm-reference-manual/blob/6f07e744cca1b900ca880e5e5251822b50fb898a/scripts/python/src/fodt/keyword_linker.py#L125

We don't link keywords in those styles

blattms (approved):

PR #474, opened 2025-01-10
Fix wrongly adjusted imports from #453

lisajulia:
In PR #453, some import statements were adjusted incorrectly


lisajulia:
@hakonhagland I'm sorry I have not noticed that when testing #453... I think I tested other commands

blattms (approved):
"grep" says that this is good.

blattms:
Maybe time to add some tests if we have spare time (famous last words :sweat_smile:)

PR #478, opened 2025-01-10
Workflow to check for missing keyword linking

hakonhagland:
See https://github.com/OPM/opm-reference-manual/pull/458#issuecomment-2583836782 for more information.


hakonhagland:
I think the workflow might be failing due to keywords with the same name appears in different chapters. For example the `ADD` keywords is defined in chapters 6, 7, 8, 9, and 10. The keyword-to-URI map is then not unique since the same keyword name can point to multiple URIs. How can we handle this?

hakonhagland:
> How can we handle this?

A quick-fix could be to always link to the first occurring section? So references to ADD would link to chapter 6, since that is the first occurring (when sorting 6, 7, 8, 9, and 10 numerically)

hakonhagland:
This is ready for review now, so I will remove the "draft state".

gdfldm:
> A quick-fix could be to always link to the first occurring section?

I think this should work for most cases. There are a few like RTEMPVD, which is in the PROPS and SOLUTION section, that is defined in the SOLUTION section.

blattms:
there is a failing [check.](/OPM/opm-reference-manual/actions/runs/12749891198/job/35533381895?pr=478) What is that about?
```

WARNING:root:Keyword name PLYDHFLF not found in bookmark content: PLY
WARNING:root:Keyword name PLYMAX not found in bookmark content: PLY
WARNING:root:Keyword name PLYVISC not found in bookmark content: PLY
WARNING:root:Keyword name PLYVISCS not found in bookmark content: PLY
WARNING:root:Keyword name PLYVISCT not found in bookmark content: PLY
WARNING:root:Keyword name SKIP100 not found in bookmark content: SKIP
WARNING:root:Keyword name SURFVISC not found in bookmark content: SURF
WARNING:root:Keyword name WLISTNAM not found in bookmark content: W
WARNING:root:Keyword name WPAVE not found in bookmark content: 

WARNING:root:Keyword name WSOLVENT not found in bookmark content: W
ERROR:root:Files have changed.
ERROR:root:Keyword SALTMF not found in old map.
Error: Process completed with exit code 1.
```

blattms:
@lisajulia Please look at the Python stuff.

hakonhagland:
> there is a failing check. What is that about?

@blattms It is this line: https://github.com/OPM/opm-reference-manual/blob/2513f966906e08368d599f1a163aa7ffd55b1401/.github/workflows/check_keyword_linking.yml#L48

I think it is nothing to worry about yet, as soon as #481 is merged, the check should pass

scripts/python/src/fodt/keyword_linker.py
```diff
@@ -438,78 +490,109 @@ def load_kw_uri_map(maindir: Path) -> dict[str, str]:
...
 #
-#   If --subsection is not given, the script will process all subsections. If --subsection
-#   is given, the script will only process the specified subsection, or if --chapter is
-#   given, the script will only process the specified chapter, or if --appendix is given,
-#   the script will only process the specified appendix. If --filename is given, the script
-#   will only process the specified file within the specified subsection.
-#   The options --appendix, --chapter and --subsection are mutually exclusive.
+#   If the option --generate-map is given, the script will generate the keyword URI map on the fly.
+#   This is useful if you suspect that libreoffice might have changed the references to the keywords.
+#   Another option is to run the script "fodt-gen-kw-uri-map" to generate a new map.
```
lisajulia, line 508: The commands `fodt-link-keywords --generate-map` and `fodt-gen-kw-uri-map` should do the same thing, no?
For me they don't:  `fodt-link-keywords --generate-map` does not change parts/meta/kw_uri_map.txt but `fodt-gen-kw-uri-map` does...

scripts/python/src/fodt/keyword_uri_map_generator.py
```diff
@@ -154,17 +161,34 @@ def add_alias(kw_uri_map: dict[str, str], keyword: str, alias: str) -> None:
...
+
 # fodt-gen-kw-uri-map
 # -------------------
 #
 # SHELL USAGE:
 #
-# fodt-gen-kw-uri-map --maindir=<main_dir> --keyword_dir=<keyword_dir>
+#  fodt-gen-kw-uri-map --maindir=<main_dir> \
+#                      --keyword_dir=<keyword_dir> \
+#                      --check-changed
```
lisajulia, line 184: Here, I ran `fodt-link-keywords --check-changed` it said: 
INFO:root:Files have not changed.
but `fodt-gen-kw-uri-map` changed said:
INFO:root:Generated keyword URI map to /home/lnebel/git_repos/opm/opm-reference-manual/scripts/python/../../parts/meta/kw_uri_map.txt

lisajulia (changes requested):

scripts/python/src/fodt/keyword_linker.py
hakonhagland: Yes, the documentation is not clear about that. I have added a commit to address this.

scripts/python/src/fodt/keyword_uri_map_generator.py
hakonhagland: I guess this is because `fodt-link-keywords --check-changed` does not include the commit in #482 ?

scripts/python/src/fodt/keyword_linker.py
lisajulia: ok, thanks!

scripts/python/src/fodt/keyword_uri_map_generator.py
lisajulia: ah sorry I just saw I was actually using the branch haha/upd_link_kw2

lisajulia: Ok, I was confused, fodt-gen-kw-uri-map said:
INFO:root:Generated keyword URI map to /home/lnebel/git_repos/opm/opm-reference-manual/scripts/python/../../parts/meta/kw_uri_map.txt
but it actually did not recreate the map.
Can you change this output?
Then I'll merge this :)

hakonhagland:
> but it actually did not recreate the map.

@lisajulia I tested it now, like this:

```
$ mv ../../parts/meta/kw_uri_map.txt /tmp
$ fodt-gen-kw-uri-map
$ ls -l ../../parts/meta/kw_uri_map.txt 
-rw-rw-r-- 1 hakon hakon 58380 Jan 22 14:23 ../../parts/meta/kw_uri_map.txt
```
so it did recreate the file here. However, it is identical to the previous version, which can be verified by running:
```
$ diff /tmp/kw_uri_map.txt ../../parts/meta/kw_uri_map.txt
# No output
```

lisajulia:
Thanks!

PR #479, opened 2025-01-13
Add missing keyword links

hakonhagland:
Reran updated keyword linking script in #478  on all files, and it detected some more missing links.


parts/chapters/subsections/12.3/DRSDTCON.fodt
```diff
@@ -4781,7 +4780,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P58">A defined character string that specifies in which cells the convective dissolution rate limit is applied, and should be set to one of the following character strings:</text:p>
        <text:list text:style-name="_40_NumberedListTable">
         <text:list-item text:start-value="1">
-         <text:p text:style-name="P104">ALL: the limit is applied to all cells.</text:p>
+         <text:p text:style-name="P104"><text:a xlink:href="#__RefHeading___Toc4420_421927891">ALL</text:a>: the limit is applied to all cells.</text:p>
```
blattms, line 4783: This and below is not used a keyword. Can/should we prevent it.

parts/chapters/subsections/12.3/DRSDTCON.fodt
```diff
@@ -4791,7 +4790,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:covered-table-cell/>
       <table:covered-table-cell/>
       <table:table-cell table:style-name="Table1966.F9" office:value-type="string">
-       <text:p text:style-name="P67">ALL</text:p>
+       <text:p text:style-name="P67"><text:a xlink:href="#__RefHeading___Toc4420_421927891">ALL</text:a></text:p>
```
blattms, line 4793: Same here

parts/chapters/subsections/8.3/EHYSTR.fodt
```diff
@@ -4726,7 +4726,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:covered-table-cell/>
       <table:covered-table-cell/>
       <table:table-cell table:style-name="Table436.F27" office:value-type="string">
-       <text:p text:style-name="P4526">OIL</text:p>
+       <text:p text:style-name="P4526"><text:a xlink:href="#__RefHeading___Toc97439_1778172979">OIL</text:a></text:p>
```
blattms, line 4729: A false positive it seems. But it is consistent with the column before (OIL, GAS as keywords even though they are parameters).

blattms (commented):
there seem to be a few false postives in this. Should we care?

Usually they don't hurt, but if you can get rid of them with little effort...

blattms:
So now we have links everywhere? :sunglasses:

hakonhagland:
> So now we have links everywhere?

@blattms Yes, I guess so :blush:

blattms:
Nice. Thanks a lot for all the hard work.

PR #481, opened 2025-01-13
Update to kw_uri_map.txt

hakonhagland:
Reran a new version of `fodt-gen-kw-uri-map` (see PR #478 for more information), also updated map to include the newly added `SALTMF` keyword.


parts/meta/kw_uri_map.txt
```diff
@@ -9,13 +9,13 @@ ACTIONW __RefHeading___Toc152225_2992482751
 ACTIONX __RefHeading___Toc152227_2992482751
 ACTNUM __RefHeading___Toc4410_421927891
 ACTPARAM __RefHeading___Toc188116_2366741785
-ADD __RefHeading___Toc99893_2023071304
-ADDREG __RefHeading___Toc99895_2023071304
+ADD __RefHeading___Toc4412_421927891
+ADDREG __RefHeading___Toc4414_421927891
```
blattms, line 13: I have the suspicion that we have broken links in the meantime. Don't we assume the bookmarks to be unchanged?

hakonhagland: > I have the suspicion that we have broken links in the meantime. 

@blattms Good observation. Yes we probably have many existing links that now are broken or linking to a wrong keyword. I have updated the script to also fix existing links according to the content of the `kw_uri_map.txt` file, see #482.

PR #482, opened 2025-01-13
Updated keyword linking script to also update the URI of existing links

hakonhagland:
Builds on #478, which should be merged first.

Updated keyword linking script to also update the URI of existing links if needed.

### DIFF
--- a/scripts/python/src/fodt/keyword_linker.py
+++ b/scripts/python/src/fodt/keyword_linker.py
@@ -87,6 +87,7 @@ class FileHandler(xml.sax.handler.ContentHandler):
         self.is_example_p = []  # Stack of boolean values: If current p tag is an example
         self.p_recursion = 0   # We can have nested p tags
         self.in_a = False
+        self.in_bookmark = False  # We should not insert links inside a bookmark
         self.in_math = False   # We should not insert links inside math tags
         self.in_binary_data = False  # We should skip binary data
         self.in_draw_frame = False  # We should not insert links in Figure captions
@@ -196,6 +197,8 @@ class FileHandler(xml.sax.handler.ContentHandler):
                 self.table_caption_info = TableCaptionInfo() # Reset the info
             elif name == "text:a":
                 self.in_a = False
+            elif name == "text:bookmark-ref":
+                self.in_bookmark = False
             elif name == "text:span":
                 self.not_keyword = False  # This cannot be nested
             elif name == "math":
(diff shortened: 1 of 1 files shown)


blattms:
THis is quite obfuscated as includes the other PR.

Can you explain what the changes to the script are?

I guess the problem of broken links arises whenever someone uses libreoffice to change a keyword file

hakonhagland:
> I guess the problem of broken links arises whenever someone uses libreoffice to change a keyword file

@blattms Yes maybe... another thing is that many of the links may not be broken, but linking to another version of the same keyword. For example, the `ADD` keyword exists in multiple chapters as noted in https://github.com/OPM/opm-reference-manual/pull/478#issuecomment-2587106692. The changes in to `kw_uri_map.txt` in #481 imposes a fixed order on those keywords, if they did not adhere to that order before, they will now need to be relinked.

lisajulia:
@hakonhagland: can you include the commit from #481 here possibly? Then this check https://github.com/OPM/opm-reference-manual/actions/runs/12757567531/job/35558097602?pr=482 should pass, right?

lisajulia:
ok, but then thick check that is failing now will only succeed once all neccessary URIs are updated and merged to master, i.e., when #483 is ready and merged, right?

hakonhagland:
> will only succeed once all neccessary URIs are updated and merged to master, i.e., when https://github.com/OPM/opm-reference-manual/pull/483 is ready and merged, right?

Yes, but I guess it will not succeed even after #483 is merged, since that is only for chapter 4, we need to also update links in chapter 5, 6, 7, ... and so on. Which will be done in separate PRs after #483 has been merged

lisajulia:
> > will only succeed once all neccessary URIs are updated and merged to master, i.e., when #483 is ready and merged, right?
> 
> Yes, but I guess it will not succeed even after #483 is merged, since that is only for chapter 4, we need to also update links in chapter 5, 6, 7, ... and so on. Which will be done in separate PRs after #483 has been merged

Ok, how about moving the addition of [this file](https://github.com/OPM/opm-reference-manual/pull/482/commits/a2c9406306fc5b8cfca6c17843e55ad0fc3cd73b#diff-f8883484598fb613431ca2e5be61feacb5964e49f5179ffc03dcc3dedce1ef30), i.e., to a separate commit and merge this when it's green? Sorry about being so picky, but I'm not a fan of merging things that make the pipeline red :thinking:

hakonhagland:
> I'm not a fan of merging things that make the pipeline red

@lisajulia I have deactivated the second check, so all checks should pass now.

PR #485, opened 2025-01-15
Update Table 2.1 in Section 2.2 for Flow 2024-10

gdfldm:
Changes made in LibreOffice


blattms:
- default linear solver is cprw. Probably also for NLDD See OPM/opm-simulators#5257
- Seems like there migth be more wrong defaults (e.g. drift compensation)

Should we do another iteration in another PR? (Current defaults may found when running without options and looking at the PRT file

gdfldm:
> * Probably also for NLDD

According to the PRT file ```NlddLocalLinearSolver="ilu0"```
and Flow --help does not list "cprw" as a option?

> * drift compensation

I have corrected ```--enable-drift-compensation```

I have generally based on the defaults on ```Flow --help-all```.  But will check against the PRT file.

blattms:
Perfect. Thanks a lot.

We developers are not really good at updating docs and help messages. :sweat_smile: That leads to confusion...

Good that we have people like you :muscle: :clap:

PR #489, opened 2025-01-24
Updated URIs in chapter 6.3

hakonhagland:


blattms:
Thanks a lot. I will not really check this...

PR #495, opened 2025-01-29
Add Highlights to Release Note

gdfldm:


blattms (approved):
Always good to see what we have improved. :muscle:

PR #496, opened 2025-01-31
Add warning about using the INCLUDE keyword in ACTIONX

lisajulia:


bska:
For what it's worth, I agree with adding this warning and the text looks good to me.

PR #504, opened 2025-02-11
Updated URIs in appendices B, C, D, and F

hakonhagland:
No updates were needed for appendix E.


parts/appendices/F.fodt
```diff
@@ -62677,7 +62676,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       </table:table-cell>
       <table:table-cell table:style-name="Table893.B2" office:value-type="string">
        <text:p text:style-name="P1536">Group, well, and connection data status keywords for this reporting time step.</text:p>
-       <text:p text:style-name="P1538">See section <text:bookmark-ref text:reference-format="number-all-superior" text:ref-name="__RefHeading___Toc204774_4221434794">A.8.4</text:bookmark-ref><text:s/><text:bookmark-ref text:reference-format="text" text:ref-name="__RefHeading___Toc204774_4221434794">RESTART Data – Group, Well and Connection Data Keywords</text:bookmark-ref>.</text:p>
+       <text:p text:style-name="P1538">See section <text:bookmark-ref text:reference-format="number-all-superior" text:ref-name="__RefHeading___Toc204774_4221434794">A.8.4</text:bookmark-ref><text:s/><text:bookmark-ref text:reference-format="text" text:ref-name="__RefHeading___Toc204774_4221434794"><text:a xlink:href="#__RefHeading___Toc135629_1317547213">RESTART</text:a> Data – Group, Well and Connection Data Keywords</text:bookmark-ref>.</text:p>
```
blattms, line 62679: This one seems wrong. It is are reference/bookmark in another one.

parts/appendices/F.fodt
```diff
@@ -62694,7 +62693,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
       <table:table-cell table:style-name="Table893.B2" office:value-type="string">
        <text:p text:style-name="P1552">Aquifer definition arrays, only applicable if an aquifer is connected to an <text:a xlink:type="simple" xlink:href="#__RefHeading___Toc55049_4106839650" text:style-name="Internet_20_link" text:visited-style-name="Visited_20_Internet_20_Link">LGR</text:a>.</text:p>
-       <text:p text:style-name="P1552">See section <text:bookmark-ref text:reference-format="number-all-superior" text:ref-name="__RefHeading___Toc195712_2189581950">A.8.6</text:bookmark-ref><text:s/><text:bookmark-ref text:reference-format="text" text:ref-name="__RefHeading___Toc195712_2189581950">RESTART Data – Aquifer Data Keywords</text:bookmark-ref>.</text:p>
+       <text:p text:style-name="P1552">See section <text:bookmark-ref text:reference-format="number-all-superior" text:ref-name="__RefHeading___Toc195712_2189581950">A.8.6</text:bookmark-ref><text:s/><text:bookmark-ref text:reference-format="text" text:ref-name="__RefHeading___Toc195712_2189581950"><text:a xlink:href="#__RefHeading___Toc135629_1317547213">RESTART</text:a> Data – Aquifer Data Keywords</text:bookmark-ref>.</text:p>
```
blattms, line 62696: Same here and  below

blattms (commented):
thanks a lot. Seems like the script somtimes adds links inside of links.

parts/appendices/F.fodt
hakonhagland: >This one seems wrong. It is are reference/bookmark in another one.

@blattms I tested this now, and it is actually allowed to have `<text:a>` inside a `<text:bookmark-ref>`. 
But it seems like the `<text:a>` will be ignored in this case. Here, is the output I get for line 62696 above after having exported to PDF:

image

so the `RESTART` keyword is not linked there. I suggest we update the script to not link keywords inside a `<text:bookmark-ref>` element.

hakonhagland: I have updated #482 to not add links inside bookmark references. Then I reran the script on appendix F, see latest commit.

PR #516, opened 2025-02-21
Typo THETA -> DTHETA

gdfldm:


parts/chapters/sections/7/2.fodt
```diff
@@ -4226,7 +4226,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       </table:table-cell>
       <table:covered-table-cell/>
       <table:table-cell table:style-name="Table822.A2" office:value-type="string">
-       <text:p text:style-name="P3641">THETA</text:p>
+       <text:p text:style-name="P3641">DTHETA</text:p>
```
blattms, line 4229: Is there a special reason why we put no link here?

gdfldm: There aren't any links in section 7.2 ?

PR #518, opened 2025-02-21
Correct equation 12.8

gdfldm:
Not sure how well this has worked. May need to double-click the equation in LibreOffice to update the image.


blattms:
The font looks strangely stretched vertically now if I open in libreoffice. Maybe use equation 12.7 as  a template

gdfldm:
I was able to get the correct size for the equation using a temporary fodt file.

blattms:
thanks. Seems like editing equations will really not work as the size if the SVG is hardcoded in XML.

Your suggestion sounds like a good approach.

PR #524, opened 2025-03-07
CO2STORE invalid change

gdfldm:


totto82:
Please dont merge this. CO2STORE support temperture above 304K. The issue reported  @blattms should not have any effect on the results. See further comments in https://github.com/OPM/opm-common/issues/4512

blattms:
There were nans, but they did not seem to present any issues

PR #526, opened 2025-03-19
Updated versions in installation notes.

blattms:


gdfldm (changes requested):
Apart from the repeated version number this looks good to me, thanks.

gdfldm:
Thanks for the update, merging now

PR #539, opened 2025-04-04
Doc the supported keywords with MICP

daavid00:
Updating the parts in the manual to align with the changes  in https://github.com/OPM/opm-common/pull/4538. In addition, writing more general 'leakage remediation' instead of 'CO2 leakage remediation'.


parts/chapters/subsections/12.3/SOURCE.fodt
```diff
@@ -4514,11 +4514,20 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
         <text:list-item>
          <text:p text:style-name="P65"><text:a xlink:href="#__RefHeading___Toc38609_2267116897">POLYMER</text:a>: the source term will inject polymer.</text:p>
         </text:list-item>
+        <text:list-item>
+         <text:p text:style-name="P65">MICRO: the source term will inject microbes.</text:p>
```
gdfldm, line 4518: @daavid00 should this be "MICR" rather than "MICRO"?

daavid00: Hi @gdfldm, thanks for reviewing the MICP PR's, and yes, it should be MICR, now it is corrected.

gdfldm:
Thank you for the updates

PR #545, opened 2025-06-11
Update Section 2.2

gdfldm:
2.2 Running OPM Flow From The Command Line


parts/chapters/2.fodt
```diff
@@ -40396,142 +40599,156 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+      <table:table-cell table:style-name="Table621.B18" office:value-type="string">
+       <text:p text:style-name="_40_Table_20_Contents">--slave</text:p>
+      </table:table-cell>
+      <table:table-cell table:style-name="Table621.C18" office:value-type="string">
+       <text:p text:style-name="_40_Table_20_Contents">A boolean value. Specify if the simulation is a slave simulation in a master-slave simulation.</text:p>
+      </table:table-cell>
+      <table:table-cell table:style-name="Table621.D18" office:value-type="string">
+       <text:p text:style-name="P127">false</text:p>
+      </table:table-cell>
+     </table:table-row>
```
blattms, line 40628: These are options that should not be used by the user. Basically flow would start processes of itself using theses for reservoir coupling (which is still not really supported). Either add a promiment note that these should never be used or skip them

blattms (changes requested):

parts/chapters/2.fodt
gdfldm: Thanks for the review. I have added a note.

blattms:
Thanks. It is not even a developer option, more of an internal one.

Issue #578, opened 2025-07-24
TUNING description


daavid00:
Item 2-2 TRGCNV is missing "target" in the keyword description. Regarding 2-7 XXXMBE, shouldn't this be green since it seems it is supported?

gdfldm:
Thanks @daavid00, please see #579. Hopefully this fixes it?

PR #579, opened 2025-07-24
Update TUNING Keyword

gdfldm:


gdfldm:
This should address issue #578

PR #581, opened 2025-07-28
review version 2025-04 Rev-A

blattms:
Actually only the updates to main.fodt in the first commit are relevant.

The rest is just cosmetics as the values get updated anyway if we open the document in libreoffice.


gdfldm (commented):
Looks good to me apart from one small nit.

PR #607, opened 2025-09-16
Partially reverting PR583

daavid00:
I was trying to export the pdf manual and I was not able to do so until I partially revert https://github.com/OPM/opm-reference-manual/pull/583 (i.e., without this PR LibreOffice will show after updating the links the app is not responding).

Ubuntu 24.04.3 LTS

LibreOffice
Version: 24.2.7.2 (X86_64) / LibreOffice Community
Build ID: 420(Build:2)
CPU threads: 12; OS: Linux 6.8; UI render: default; VCL: gtk3
Locale: en-US (en_US.UTF-8); UI: en-US
Ubuntu package version: 4:24.2.7-0ubuntu0.24.04.4
Calc: threaded

Does anyone else with that version of Ubuntu and LibreOffice manage to build the manual? 
I still cannot understand the issue.

### DIFF
--- a/parts/appendices/E.fodt
+++ b/parts/appendices/E.fodt
@@ -21511,7 +21511,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P674">--ilu-redblack</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table4.C53" office:value-type="string">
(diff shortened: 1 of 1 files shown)


daavid00:
Thanks @gdfldm for looking at this, and happy to hear that LO on Win11 works fine. If I am the only one with this issue, I will close the PR, and at least now I have a work around.

PR #622, opened 2025-10-15
EHYSTR Keyword add capillary pressure correction option

gdfldm:
item 14


totto82:
@Edmund provided some references and notes on the correction. Should this be reflected in the manual?

https://github.com/OPM/opm-simulators/issues/6383#issuecomment-3329954105

blattms:
@gdfldm Will add something here or should we merge?

gdfldm:
> Will add something here or should we merge?

I have added a link to the discussion. I think this would be ok for now since this extra item may only be a temporary work around?

PR #623, opened 2025-10-16
Updating BIOFILM and MICP documentation

daavid00:
Following https://github.com/OPM/opm-common/pull/4733, this PR documents the `BIOFILM` model, as well as updating links and text in the common keywords also used in `MICP`.

### DIFF
--- a/parts/chapters/subsections/12.3/WMICP.fodt
+++ b/parts/chapters/subsections/12.3/WMICP.fodt
@@ -4395,7 +4395,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:covered-table-cell/>
       <table:covered-table-cell/>
       <table:table-cell table:style-name="Table2030.F4" table:number-rows-spanned="2" office:value-type="string">
-       <text:p text:style-name="P7542">None</text:p>
+       <text:p text:style-name="P7542">0.0</text:p>
       </table:table-cell>
      </table:table-row>
      <table:table-row table:style-name="Table2030.1">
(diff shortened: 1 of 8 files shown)


parts/chapters/subsections/8.3/BIOFPARA.fodt
```diff
@@ -4637,7 +4637,7 @@ MICP update.</dc:description><meta:initial-creator>David Landa-Marbán</meta:ini
        <text:p text:style-name="_40_Table_20_Contents">DETAEXPO</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table2011.C2" table:number-columns-spanned="3" office:value-type="string">
-       <text:p text:style-name="_40_Table_20_Contents">A real value that defines the exponent in the norm for the water velocity in the detachment term.</text:p>
+       <text:p text:style-name="_40_Table_20_Contents">A real value that defines the exponent in the norm for the water velocity in the detachment term (this value was set to 0.58 in the MICP publications).</text:p>
```
gdfldm: Looks like you need a link here

parts/chapters/subsections/12.3/WMICP.fodt
gdfldm: The keyword definition says that there are no default values for the concentrations. So should the example have zero values rather than defaulted concentrations?

gdfldm:
@daavid00, thanks for the update. Apart from the comments above this all looks good to me.

daavid00:
Thanks @gdfldm, now your comments have been included.

PR #654, opened 2025-11-19
supporting positive EQLOPT3

ElyesAhmed:
Preparing the manual for EQLOPT3 positive to be used in EQUIL(ELOPT3): this corresponds to item 9 for the target accuracy method (N>0)


blattms:
Thanks a lot for your contribution. We highly value it.

Is this for release 2025.10 or the next one?

Unfortunately, it seems like you might have directly edited a libreoffoce file (10.fodt) that includes other files. The problem with this is that all included files become part of that one when you save. This breaks our structure and we probably cannot merge this as is.

I tried to persuade libreoffice to tell me the text changes, but it failed.

If it is feasible for you then the easiest way is to edit the *fodt files directly with an editor (and not libreoffice). Or suggest your text changes here or in an issue in clear text.

Another option is to tell us which sections or subsections you have changed. Then @hakonhagland could make sure that they in a separate file that does not include other files. The I think you could use libreoffice on those and later we could see the changes by comparing files.

I am sorry for the inconvenience.

gdfldm:
This has PR has been replaced by #655.

PR #686, opened 2025-12-08
Section 2.4.2: Update INFOSTEP and INFOITER

### LINKED ISSUE #677: INFOITER and INFOSTEP updates
Figures 2.35 (INFOSTEP Report Example) and 2.36 (INFOITER Report Example) need to be updated, since now INFOSTEP has an extra column (LocSol), and INFOITER a few more (e.g., CnvPvFracConv, PenaltyDecay, etc.). Then the text in this section (2.4.2 Convergence and Numerical Performance) should be also updated.

gdfldm:
Closes #677


blattms (changes requested):
Thanks a lot.

I think we need to describe what a local solver is. I guess this is the solver within NLDD and hence with it it is zero? @atgeirr please propose something

In Figure 36 one cannot anything anymore, because the font is so small. Can we dosomething about this? Maybe include the picture in landscape mode?

PR #709, opened 2026-04-08
Fix Equation 2.4

### LINKED ISSUE #708: a typo for the parameter is the maximum water relative permeability
I have notice there may be a typo in this section for the maximum water relative permeability in P121 (OPM FLOW REFERENCE MANUAL (2025-10) Revision: Rev-0). There is missing a little w in the parameter I highlighted in the attachment for your reference.

Image

gdfldm:
krww -> krw

Fixes #708


parts/chapters/2.fodt
michal-toth, line 74580: Should this be "water relative permeability _at_ Sorw" instead of 'st' Sorw?

michal-toth:
> krww -> krw
> 
> Fixes #708

The #708 mentions a missing w in the description of the parameter krww (which is mentioned as krw). This PR changes krww to krw, Swc to Swcr, and krww to krw. The problem in #708 does not seem to be addressed.

krw and krww are two different variables. Please double-check the change from krww to krw too.

PR #715, opened 2026-04-09
GCONPROD Keyword add support for Item 10 = POTN

gdfldm:
and update text


parts/chapters/subsections/12.3/GCONPROD.fodt
```diff
@@ -5761,7 +5761,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+    <text:p text:style-name="P127">All groups are controlled by oil rate targets or constraints, but only the field level has water, gas and liquid constraints to reflect the actual production facility constraints. The wells under group control will be produced based on oil potential of each of the wells under group control, such that the field oil production target of 40,000 stb/d is honored and subject to the other constraints. <text:s/>In addition, GRP01 and GRP02 oil rate values of 25,000 stb/d are constraints as these two groups are subject to the <text:span text:style-name="NotKeyword">FIELD</text:span> level targets and constraints.</text:p>
```
michal-toth: This sentence should use active voice: "The wells under group control will _be produced_ based on oil potential..."

gdfldm: Changed to "The wells under group control are produced by the group based on the oil potential..."

michal-toth: I meant that the sentence should be "The wells under group control will _produce_ based on oil potential...". What is produced is not wells but what gets out of them.

michal-toth (approved):

PR #724, opened 2026-04-25
Fix typos and unit formatting

michael-liem-4:
Fix minor typos in Table 5.2 and in the THPRESFT keyword description. Fix unit formatting in WCONINJE and WCONINJH keyword descriptions.


gdfldm:
@michael-liem-4, thanks for the fixes.

PR #732, opened 2026-04-28
Fix degree symbol

gdfldm:
Use degree symbol rather than superscript o


parts/chapters/sections/8/2.fodt
```diff
@@ -10949,7 +10949,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+    <text:p text:style-name="_40_TextBody">The term low shrinkage oil is derived from the fact that oil shrinkage is small, i.e., with an oil formation volume factors less than 1.5. This oil can be described as having a broad based phase envelope, high percentage of liquid, high proportion of heavier hydrocarbons, GOR&apos;s less than 500 scf/stb or 100 Sm<text:span text:style-name="T176">3</text:span>/m<text:span text:style-name="T176">3</text:span>, oil gravity 30 °<text:span text:style-name="NotKeyword">API</text:span> or heavier, and the stock tank liquid being black or a deep color. If the oil has a very low GOR and therefore has an oil formation volumes factor close to one, as observed in heavy oil reservoirs, then it is common to model this type of reservoir neglecting the gas phase. That is this type of reservoir fluid is typically modeled as a two-phase oil-water system using a black-oil formulation, which results in greater computational efficient due to having only two phases as opposed to the three phases in an oil-gas-water system.</text:p>
```
michal-toth: Old typo: "which results in greater computational _efficient_ due to".

The style T176 could be replaced with Super too.

parts/chapters/subsections/10.3/OILAPI.fodt
```diff
@@ -4295,7 +4295,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
...
+       <text:p text:style-name="P4573">The American Petroleum Institute (“API”) classifies oils based on an <text:a xlink:href="#__RefHeading___Toc4422_421927891">API</text:a> gravity (γ<text:span text:style-name="T4092"><text:a xlink:href="#__RefHeading___Toc4422_421927891">API</text:a></text:span>), <text:s/>or degrees <text:a xlink:href="#__RefHeading___Toc4422_421927891">API</text:a> (°<text:span text:style-name="NotKeyword">API</text:span>), the relationship between relative density (γ<text:span text:style-name="T4092">o</text:span>) of oil and <text:a xlink:href="#__RefHeading___Toc4422_421927891">API</text:a> gravity (γ<text:span text:style-name="T4092"><text:a xlink:href="#__RefHeading___Toc4422_421927891">API</text:a></text:span>) is given by:</text:p>
```
michal-toth, line 4298: The API with the degree symbol lost the xlink. However, there are many other API around with links.

michal-toth:
Looks good. It is unfortunate that using the degree symbol broke xlinks to many API.

michal-toth (approved):

gdfldm:
> Looks good. It is unfortunate that using the degree symbol broke xlinks to many API.

I guess this is subjective. I don't think it always makes sense to generate a link if it is not relevant or necessary, for example if FIELD refers to the top level group it doesn't make sense to create a link to the FIELD units keyword, or if there are multiple instances in the same paragraph.

The "NotKeyword" tag has been used here to manually prevent links from being generated by the script.

PR #736, opened 2026-05-06
Typo 'as oppose to' -> 'as opposed to'

gdfldm:


michal-toth (approved):
It is great that you checked other files for the typo.

PR #741, opened 2026-05-08
Issue 708: Update equations 8.19 and 8.20

gdfldm:


michal-toth (approved):
Good catch finding the typo k_rw -> k_rw,max (should be k_rww before).

By chance, I found another typo in this file: "Under these circumstance" (at two places).

michal-toth (approved):

PR #747, opened 2026-05-12
Fix floating point exponents

gdfldm:
and tidy FOAMFSC keyword expressions at the same time

### DIFF
--- a/parts/chapters/subsections/12.3/GCONCAL.fodt
+++ b/parts/chapters/subsections/12.3/GCONCAL.fodt
@@ -4538,7 +4538,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       </table:table-cell>
       <table:table-cell table:style-name="Table1144.C2" table:number-columns-spanned="3" office:value-type="string">
        <text:p text:style-name="P38">A real positive value that defines the target surface calorific value of the produced gas.</text:p>
-       <text:p text:style-name="P34">The default value of 1 x 10<text:span text:style-name="Super">20</text:span> switches off calorific control for the group.</text:p>
+       <text:p text:style-name="P34">The default value of 10<text:span text:style-name="Super">20</text:span> switches off calorific control for the group.</text:p>
       </table:table-cell>
       <table:covered-table-cell/>
       <table:covered-table-cell/>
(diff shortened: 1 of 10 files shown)


parts/appendices/F.fodt
```diff
@@ -44891,7 +44894,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P1561">Data</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table783.C5" table:number-columns-spanned="3" office:value-type="string">
-       <text:p text:style-name="P1562">Tabular data array. The data should be written out in double precision and missing data should be set to less than or equal to -10.0 x 1020 or greater than or equal to 10.0 x 1020.</text:p>
+       <text:p text:style-name="P1562">Tabular data array. The data should be written out in double precision and missing data should be set to less than or equal to -10.0 x 10<text:span text:style-name="Super">20</text:span> or greater than or equal to 10.0 x 10<text:span text:style-name="Super">20</text:span>.</text:p>
```
michal-toth: With proper exponents it is now better to write 10^21 than 10x10^20.

parts/chapters/subsections/12.3/GCONCAL.fodt
```diff
@@ -4535,7 +4538,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       </table:table-cell>
       <table:table-cell table:style-name="Table1144.C2" table:number-columns-spanned="3" office:value-type="string">
        <text:p text:style-name="P38">A real positive value that defines the target surface calorific value of the produced gas.</text:p>
-       <text:p text:style-name="P34">The default value of 1 x 1020 switches off calorific control for the group.</text:p>
+       <text:p text:style-name="P34">The default value of 1 x 10<text:span text:style-name="Super">20</text:span> switches off calorific control for the group.</text:p>
```
michal-toth: 1x in front of 1x10^20 is no longer necessary

parts/appendices/F.fodt
gdfldm: I suspect that this is a typo and 10 x 10^20 should be 1 x 10^20 [see for example](https://github.com/OPM/opm-common/blob/fbb93414dfcc5e1a19a642fd34306af76f4137f2/opm/output/eclipse/WriteInit.cpp#L592). I will change it

michal-toth:
There are couple of 1x10^n and 1.0x10^n left. Everything else looks good.

gdfldm:
> There are couple of 1x10^n and 1.0x10^n left. Everything else looks good.

I am undecided about the best format to use: 10^n, 1 x 10^n, or 1.0 x 10^n, or even 1en, 1En, or 1E+n for default values. So I am minded to leave these for now and consider a possible global change in a future PR.

michal-toth:
> I am undecided about the best format to use: 10^n, 1 x 10^n, or 1.0 x 10^n, or even 1en, 1En, or 1E+n for default values. So I am minded to leave these for now and consider a possible global change in a future PR.

Then this PR can be merged. I prefer 10^n for integers but am less certain about floating point numbers.

michal-toth (approved):

PR #753, opened 2026-05-14
Update Radial Grid Keywords ...

gdfldm:
since OPM ResInsight now supports plotting radial grids.


michal-toth (approved):
Nice to see that ResInsight can now process radial grids. The PR looks good, although I am think there is an old typo. The word "whereas," should not be followed by a comma. (It is in every paragraph changed by this PR.)

michal-toth (approved):

PR #760, opened 2026-05-26
Appendix C: Fix page break

gdfldm:
Fix missing page break before Appendix C.
(changes based on B.fodt)


michal-toth:
Where is the page break that got fixed? The only change I found is the formatting of the first/title page of the appendix. Adding the background coloring, headers, etc. is a nice improvement (and approved :) ).

gdfldm:
> Where is the page break that got fixed?

The missing page break only became apparent when the full manual was built. I think previously the page break was at the end Appendix B instead of at the start of Appendix C, and I deleted it when I updated the Release Note.

I have now added a page break at the start of Appendix C, which is now more consistent with the other appendices. The background colours etc. are a side effect of moving the page break.

michal-toth (approved):

Issue #14, opened 2023-12-13
Units missing in salt keywords explanation


gdfldm:
Units missing in salt keywords explanation

PR #761, opened 2026-05-28
Minor typos and corrections #2

pkirkham:
Second small batch of minor typos and corrections noticed in the manual.


pkirkham:
Also addresses issue #14. Seemed trivial to add some words so this can be closed.

parts/chapters/subsections/12.3/WELSPECS.fodt
```diff
@@ -5016,7 +5016,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
        <text:p text:style-name="P135">WELNAME</text:p>
       </table:table-cell>
       <table:table-cell table:style-name="Table665.C2" table:number-columns-spanned="3" office:value-type="string">
-       <text:p text:style-name="P135">A character string of up to eight characters in length that defines the well name for which the well specification data are being defined.</text:p>
+       <text:p text:style-name="P135">A character string of up to eight characters in length that defines the well name for which the well specification data are being defined. Can be set to '?' to apply to any well that matches the condition within an <text:a xlink:href="#__RefHeading___Toc152227_2992482751">ACTIONX</text:a> block.</text:p>
```
gdfldm: @pkirkham thank you for all the fixes. I would prefer not to add this here (since this is described onthe ACTIONX page and a change here would suggest a change to all well keywords). But the well name description for all well keywords should probably be updated to include well lists and wildcards in a future PR.

parts/chapters/subsections/8.3/EHYSTR.fodt
```diff
@@ -4515,7 +4515,7 @@ Updated with AFR/TSA Rev-D comments and new keywords.</dc:description><meta:init
       <table:covered-table-cell/>
       <table:covered-table-cell/>
       <table:table-cell table:style-name="Table436.C27" table:number-columns-spanned="3" office:value-type="string">
-       <text:p text:style-name="P4517">Oil Wet to Water Wet Water Wet </text:p>
+       <text:p text:style-name="P4517">Oil Wet to Water Wet </text:p>
```
gdfldm: How about "Oil Wet to Water Hysteresis Models"

parts/chapters/subsections/12.3/WELSPECS.fodt
pkirkham: Not including on just one keyword, but considering for future implementation on multiple keywords makes sense to me.

parts/chapters/subsections/8.3/EHYSTR.fodt
pkirkham: I think that reads better in the table. I was simply removing the obvious typo, but adding "Hysteresis Models" is more consistent with the row above in that table. I agree with your suggested wording. Should I resubmit a PR or will you make the change directly?

pkirkham: I found I could edit the submitted text directly in my fork. I think this should now appear in the same PR?

parts/chapters/subsections/12.3/WELSPECS.fodt
pkirkham: I think I've removed the proposed change from my PR. This is my first time doing this so apologies if I'm doing the wrong thing.

gdfldm:
Thank you for making the changes. That all looks good to me.

